PR #22, opened 2019-06-18
Update Unity project settings to be able to build the UWP solution for HoloLens

matthejo:
Main changes were:
1. gitignore updates to mask out generated content that results from updating asset caches and building the UWP solution
2. Updating the capabilities and package name (specifically the name of the project can't end with .Unity, as this causes a namespace conflict in the generated App.cs in the UWP solution)
3. Removing a missing GazeProvider component in SpectatorView.ArUco.HoloLens

### DIFF
--- a/src/SpectatorView.Unity/.gitignore
+++ b/src/SpectatorView.Unity/.gitignore
@@ -61,6 +61,19 @@ sysinfo.txt
 # Crashlytics generated file
 crashlytics-build.properties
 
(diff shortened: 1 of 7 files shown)


chrisfromwork:
Are these changes to manifest.json, GraphicSettings.asset and EditorBuildSettings.asset really needed? I think we should avoid introducing a scene into the EditorBuildSettings.asset definition for now.

PR #28, opened 2019-06-20
Enable qr code marker visuals + qr code marker detectors

chrisfromwork:
This review adds the following functionality:

1) SpectatorView.cs is updated to take different debug visual prefabs for users/spectators. This makes it easier to understand who observed coordinates at what location.
2) An Axis2 prefab was created for usage on the spectator device.
3) Some scenes were moved around, all scenes were updated to contain the new spectator/user debug visual prefabs

4) MarkerVisualSpatialLocalizer and MarkerVisualDetectorSpatialLocalizer were changed to abstract classes. Separate implementations were created for ArUco vs QR codes so that different localizer ids could be used for the different marker detector experiences.
5) MarkerVisualDetectorLocalizationInitializer was removed. Now, MarkerVisualLocalizationInitializer instructs the user device on what detector to use.
6) MarkerVisualSpatialLocalizers now take prefabs containing IMarkerVisuals. This avoids requiring the IMarkerVisual related ui to remain in the scene when unneeded.
7) ArUco and QR Code marker visual spatial localizers were all added to the main spectator view spatial localizers prefab.

Related to issue: https://github.com/microsoft/MixedReality-SpectatorView/issues/8


src/SpectatorView.Unity/Assets/MarkerDetection/IMarkerVisual.cs
```diff
@@ -34,5 +34,12 @@ public interface IMarkerVisual
         /// <param name="supportedIds">Output maximum marker id</param>
         /// <returns>Returns true if the maximum marker id can be found, otherwise false.</returns>
         bool TryGetMaxSupportedMarkerId(out int markerId);
+
+        /// <summary>
+        /// Tries to set an additional scale factors for the marker visual.
```
matthejo: >Tries to set an additional scale factors for the marker visual. [](start = 12, length = 63)

I think more  description might be necessary here. How does this differ from the marker size?

src/SpectatorView.Unity/Assets/SpectatorView/Scenes/EditorHoloLens2ExtrinsicsCalibrationScene.unity.meta
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: dacd2c67aa5f55a4bb0f866db0bbb7a4
+guid: fca241a083bd79b4c8c1a8199e4569d9
```
matthejo: >fca241a083bd79b4c8c1a8199e4569d9 [](start = 6, length = 32)

Why did several of these meta files change?

PR #53, opened 2019-06-27
Update project settings to have consistent package names across platforms

chrisfromwork:
This change adds consistent package names across platforms. Previously iOS had nothing set.


chrisfromwork:
This addresses https://github.com/microsoft/MixedReality-SpectatorView/issues/23

PR #69, opened 2019-07-02
Fix compile warnings

chrisfromwork:
This change fixes compile warnings for SpectatorView code. ASA compile warnings still exist.


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Calibration/CompositorWrapper.cs
```diff
@@ -12,7 +12,11 @@ internal class CompositorWrapper : Singleton<CompositorWrapper>
         /// </summary>
         [Tooltip("CompositionManager used to obtain the DSLR stream")]
         [SerializeField]
+#if UNITY_EDITOR
         private CompositionManager compositionManager = null;
+#else
+        private CompositionManager compositionManager;
+#endif
```
matthejo: Is there a reason this can't have the "= null" always?

chrisfromwork: it generates warnings based on whether the preprocessor directive is/isn't enabled.

andreiborodin: It's the bizarre combination of needing the serializefield, but this class usage is editor only

PR #72, opened 2019-07-03
Add UI in the Compositor window to choose a type of spatial localizer and configure it

matthejo:
This change adds new UI in the Compositor window for choosing a type of spatial localizer and for editing that localizer's settings.

Supporting settings editing is optional for a localizer. If the object implementing ISpatialLocalizationSettings also implements IEditableSpatialLocalizationSettings, that implementation will be used to present content in a Unity popup window.

The list of supported localizers is determined by cross-referencing the connected device's supported localizers with the compositor's supported localizers.

Here is an example of selecting a supported ArUco Marker localizer and configuring its settings.
image


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/ISpatialLocalizationSettings.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.IO;
+using UnityEditor;
```
chrisfromwork: Is this needed? Can it be removed?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/MarkerVisual/MarkerVisualSpatialLocalizer.cs
```diff
@@ -6,6 +6,7 @@
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using UnityEditor;
 using UnityEngine;
```
chrisfromwork: Is this needed? nit: please revert this file if no changes occurred.

matthejo: I think these were remnants from an intermediate change, will revert them.

PR #110, opened 2019-07-09
Add known issue for ios screen recording

chrisfromwork:
This change is to address: https://github.com/microsoft/MixedReality-SpectatorView/issues/109


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/ScreenRecording/README.md
```diff
@@ -7,6 +7,12 @@ Spectator View supports filming HoloLens experiences with mobile devices. All mo
 
 >Note: On Android, recorded content is always saved to a file; whereas on iOS, recorded content is saved to a temporary file that must be manually saved by the user.
 
+### Known Issues
+#### __Issue:__ iOS screen recording fails to create a video
```
matthejo: It feels like this should go at the end of the document, and maybe be called something like "Troubleshooting". "Known Issues" makes it sound like a bug in SpectatorView that just hasn't been fixed yet.

Issue #111, opened 2019-07-10
QRTracker fails to initialize correctly on clean install for QRCodeMarkerDetector


chrisfromwork:
It seems the QRTracker needs to be created after requesting the webcam capability.

PR #112, opened 2019-07-10
Fix qr code marker detection to initialize correctly on a clean install

chrisfromwork:
This review addresses https://github.com/microsoft/MixedReality-SpectatorView/issues/111

1) A DebugLog function is added to QRCodeMarkerDetector to cut down on log noise.
2) Capability requests for the webcam capability are relocated to the QRCodesManager and are called prior to creating the QRTracker. It appears that the capability needs to be granted prior to creating the QRTracker compared to before starting the QRTracker.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
@@ -8,8 +8,9 @@
 
 using UnityEngine;
 
-using System.Collections.Generic;
 using System;
+using System.Threading.Tasks;
+using System.Collections.Generic;
 
 #if ENABLE_QRCODES
 using Windows.Perception.Spatial;
@@ -24,6 +25,10 @@ namespace Microsoft.MixedReality.SpectatorView
     public class QRCodeMarkerDetector : MonoBehaviour,
         IMarkerDetector
     {
+        [Tooltip("Check to enable debug logging")]
+        [SerializeField]
+        private bool debugLogging = false;
+
 #if ENABLE_QRCODES
         private QRCodesManager _qrCodesManager;
         private Dictionary<Guid, SpatialCoordinateSystem> _markerCoordinateSystems = new Dictionary<Guid, SpatialCoordinateSystem>();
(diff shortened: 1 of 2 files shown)


andreiborodin:
> // Copyright (c) Microsoft Corporation. All rights reserved.

Why are these files so different that neither tool can diff them properly?

---
Refers to: src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodesManager.cs:1 in 0b0fb62. [](commit_id = 0b0fb626ce215571d9aadbddb3da7cee06beb08f, deletion_comment = False)

chrisfromwork:
> // Copyright (c) Microsoft Corporation. All rights reserved.

There were inconsistent line endings that I switched to windows.

---
In reply to: [510221931](https://github.com/microsoft/MixedReality-SpectatorView/pull/112#issuecomment-510221931) [](ancestors = 510221931)

---
Refers to: src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodesManager.cs:1 in 0b0fb62. [](commit_id = 0b0fb626ce215571d9aadbddb3da7cee06beb08f, deletion_comment = False)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
```diff
@@ -88,26 +93,43 @@ public bool TryGetMarkerSize(int markerId, out float size)
...
+            {
+                StopTracking();
+                _qrCodesManager.QRCodeAdded -= QRCodeAdded;
+                _qrCodesManager.QRCodeRemoved -= QRCodeRemoved;
+                _qrCodesManager.QRCodeUpdated -= QRCodeUpdated;
+            }
         }
 
-        private async void StartTracking()
+        private async Task StartTracking()
```
matthejo: >StartTracking [](start = 27, length = 13)

Nit: I prefer the naming convention where async methods have a name that ends in Async (StartTrackingAsync). Makes it a lot easier when reading code to tell if the method is synchronous or not

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodesManager.cs
```diff
@@ -1,302 +1,316 @@
...
+
+        protected async void Start()
+        {
+            if (AutoStartQRTracking)
+            {
+                await StartQRTracking();
+            }
+        }
+
+        public async Task<QRTrackerStartResult> StartQRTracking()
```
matthejo: >StartQRTracking [](start = 48, length = 15)

It looks like if you call StartQRTracking (which is async) and then call StopQRTracking before the "await cap.RequestAccessAsync" finishes, the end result would be that the tracker would be running.

I think to make this safe to that kind of timing, you need to separate the state of "IsTrackerRunning" from "ShouldTrackerBeRunning". Then StopQRTracking can always set a state of ShouldTrackerBeRunning to false, and StartQRTracking can check for ShouldTrackerBeRunning after any async call it makes to ensure that it should continue to run the rest of the start sequence

chrisfromwork:
I'll address this in https://github.com/microsoft/MixedReality-SpectatorView/pull/120

PR #119, opened 2019-07-11
Allow external scripts to choose connecting to or listening on non-default ports

matthejo:
For internal demos, we sometimes set up connections on ports other than the default-configured port for a NetworkManager. This change allows an external script to connect to a port other than the RemotePort for a NetworkManager, and also allows listening on a port other than RemotePort.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
@@ -33,13 +33,22 @@ namespace Microsoft.MixedReality.SpectatorView
         /// </summary>
         protected abstract int RemotePort { get; }
 
-        /// <summary>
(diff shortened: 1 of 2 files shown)


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
```diff
@@ -33,13 +33,22 @@ public abstract class NetworkManager<TService> : CommandRegistry<TService>, INet
...
         protected abstract int RemotePort { get; }
 
-        /// <summary>
-        /// Connects to the holographic camera rig with the provided remote IP address.
-        /// </summary>
-        /// <param name="remoteAddress">The IP address of the holographic camera rig's HoloLens.</param>
+        /// <inheritdoc />
+        public void StartListening(int port)
+        {
+            connectionManager.StartListening(port);
```
chrisfromwork: nit: should we check whether connectionManager is null?

PR #120, opened 2019-07-15
DSLR Calibration cleanup

chrisfromwork:
This review contains the following changes:

1) pdf and png files for the Chessboard and Calibration Board used for DSLR calibration.
2) Updated documentation that explains the DSLR calibration process
3) A new CalibrationWindow component that provides richer information to the user performing DSLR calibration
4) Calibration logic has been added to the HolographicCamera scene. This reduces the number of applications that need to run on the Rig mounted HoloLens 2 device.
5) Calibration logic has been updated to use the TCPConnectionManager and INetworkManager classes. This has resulted in the ability to remove deprecated/legacy interfaces.
6) The ability to undistort chessboard images using calculated camera intrinsics was added to the native SpectatorView.OpenCV plugin.


doc/SpectatorView.Setup.DSLR.md
```diff
@@ -1 +1,164 @@
...
+
+> Note: DSLR Filming currently requires a HoloLens 2 for the camera rig. HoloLens 1 devices are not supported.
+
+
+1. HoloLens 2 (This device will be needed in addition to the device worn by the user)
+2. Windows PC
+3. Visual Studio 2017 installed on the PC
+4. Windows 10 SDK (10.0.18362.0)
+5. Unity installed on the PC
+6. Digital single-lens reflex (DSLR) camera
```
matthejo: >Digital single-lens reflex (DSLR) camera [](start = 3, length = 40)

I think technically any video camera that can output 1080p video over SDI or HDMI will work, the sensor type is not important.

doc/SpectatorView.Setup.DSLR.md
```diff
@@ -1 +1,164 @@
...
+    6. Run the SpectatorViewCompositor scene in the Unity Editor.
+    7. You should see the Camera Feed appear in the Compositor window if everything has been configured correctly.
+
+13. Attach your HoloLens 2 device to the DSLR Camera using the HoloLens 2 camera mount (Coming soon...)
+
+## Camera Intrinsics
+
+Camera intrinsics quantify focal lengths, principal points and lens distortion information for your DSLR camera. For DSLR filming, camera intrinsics are used to calculate the Unity camera's projection matrix for the hologram feed. This then allows the hologram feed to be composited with the DSLR camera feed. For more information on camera intrinsics, see [here](https://en.wikipedia.org/wiki/Camera_resectioning).
+
+>Note: Changing the zoom and focus length of your DSLR camera's lens will change the camera intrinsics. When conducting calibration, you should change your lens to its manual focus setting. You should also avoid using the manual focus ring to change focus when obtaining images. Although images may at times appear blurry, as long as the are kept at a consistent focus length, you should be able to obtain a usable camera intrinsics.
```
matthejo: >Although images may at times appear blurry, as long as the are kept at a consistent focus length, you should be able to obtain a usable camera intrinsics. [](start = 279, length = 154)

I think it might be worth adding something like this: "For best results, use a bright external light to illuminate your calibration board. Set your lens's aperture to the smallest aperture (largest f-stop value) for which images still are exposed enough to see the marker pattern. Smaller apertures increase the depth of field, and will allow for less focus blur when capturing images at different depths."

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/HeadsetRequestHandler.cs
```diff
@@ -0,0 +1,193 @@
...
+                updateData = true;
+            }
+            else
+            {
+                Debug.LogWarning("Received network payload that wasn't a headset calibration data request");
+                request = null;
+            }
+        }
+
+#if WINDOWS_UWP
```
matthejo: >WINDOWS_UWP [](start = 4, length = 11)

I know in the past I've had problems with WINDOWS_UWP being defined at weird parts of the Unity build process, where UNITY_WSA is always defined at the correct times. I'd double check that every build configuration (.NET Scripting + IL2CPP) works with this define, or switch it to UNITY_WSA

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Calibration/EditorExtrinsicsCalibration.cs
```diff
@@ -213,24 +175,55 @@ private void Update()
...
+
+        public void RequestHeadsetData()
+        {
+            if (holographicCameraObserver != null &&
+                holographicCameraObserver.IsConnected)
+            {
+                using (MemoryStream memoryStream = new MemoryStream())
+                {
+                    using (BinaryWriter writer = new BinaryWriter(memoryStream))
+                    {
```
matthejo: Nit: you can place both of these using blocks together without each needing their own brace block, which is pattern used extensively elsewhere in the project.

using (MemoryStream memoryStream = new MemoryStream())
using (BinaryWriter writer = new BinaryWriter(memoryStream))
{
   ...
}

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodesManager.cs
```diff
@@ -1,302 +1,337 @@
...
+            if (AutoStartQRTracking)
+            {
+                await StartQRTrackingAsync();
+            }
+        }
+
+        public async Task<QRTrackerStartResult> StartQRTrackingAsync()
+        {
+            bool attemptToStart = false;
+            lock (startupLock)
```
matthejo: >lock (startupLock) [](start = 12, length = 18)

Is this lock necessary? I think it's only necessary if something calls either of these methods from a different thread. If they're called from the Unity main thread, then the lock isn't necessary. Even after the awaited method returns the execution will still be on the Unity main thread due to its synchronization context.

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CalibrationWindow.cs
```diff
@@ -0,0 +1,185 @@
...
+            {
+                GUILayout.BeginHorizontal();
+                {
+                    holographicCameraIPAddress = EditorGUILayout.TextField(holographicCameraIPAddress);
+                    ConnectButtonGUI(holographicCameraIPAddress, cameraDevice);
+                }
+                GUILayout.EndHorizontal();
+
+                GUILayout.Label(notConnectedMessage);
+            }
```
matthejo: I think this is almost entirely copy-pasted from the base class, can we extract this into its own method in the base class and then you could just call that method from here? I'm thinking it could be called something like "ConnectionStatusGUI"

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodesManager.cs
chrisfromwork: @andreiborodin and I chatted offline and he provided some great perspectives on how to handle this locking better. The code should now be updated.

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/HeadsetRequestHandler.cs
```diff
@@ -0,0 +1,184 @@
...
+                    child.gameObject.SetActive(false);
+                }
+
+                initialized = false;
+            }
+        }
+
+        private void OnDisconnect(SocketEndpoint endpoint)
+        {
+            if(endpoint.Address == editorAddress)
```
andreiborodin: >( [](start = 14, length = 1)

nit:spacing

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/HeadsetRequestHandler.cs
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        private void OnHeadsetCalibrationUpdated(HeadsetCalibrationData data)
+        {
+            if (holographicCameraBroadcaster != null)
+            {
+                using (MemoryStream memoryStream = new MemoryStream())
+                {
+                    using (BinaryWriter writer = new BinaryWriter(memoryStream))
+                    {
```
andreiborodin: You can write these as:
using (...)
using (...)
{
}

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/HeadsetRequestHandler.cs
chrisfromwork: Tested for both .net and il2cpp

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Calibration/CalculatedCameraCalibration.cs
```diff
@@ -2,26 +2,52 @@
...
+    /// <summary>
+    /// A class containing camera intrinsic and extrinsic information produced through calibration.
+    /// </summary>
     [Serializable]
     public class CalculatedCameraCalibration
     {
+        /// <summary>
+        /// Camera extrinsics calculated through calibration.
+        /// </summary>
         public CalculatedCameraExtrinsics Extrinsics;
```
andreiborodin: Make these properties

andreiborodin: >public CalculatedCameraExtrinsics Extrinsics; [](start = 8, length = 45)

Can these be set externally?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodesManager.cs
```diff
@@ -1,302 +1,387 @@
...
+                stopTrackerTask = null;
+                startTrackerCTS = new CancellationTokenSource();
+
+                var hybridCTS = CancellationTokenSource.CreateLinkedTokenSource(startTrackerCTS.Token, cancellationToken);
+                DebugLog("Kicking off a new start tracker task");
+                return startTrackerTask = Task.Run(() => StartQRTrackingAsyncImpl(hybridCTS.Token), hybridCTS.Token);
+            }
+        }
+
+        private async Task<QRTrackerStartResult> StartQRTrackingAsyncImpl(CancellationToken token)
```
matthejo: >token [](start = 92, length = 5)

Should this cancellation token be used anywhere? A lot of work is going in to creating the right kind of linked token to never use it :-)

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scenes/SpectatorView.ExtrinsicsCalibration.unity
```diff
@@ -473,11 +426,9 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 7abda130a4f4dfc4989cb445d1bd8fae, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  cameraIntrinsicsPath: C:\Users\chriba\Documents\Calibration\CameraIntrinsics_1.json
-  expectedNumberOfMarkers: 18
-  MatchMakingService: {fileID: 740770235}
-  NetworkingService: {fileID: 740770235}
-  headsetCalibration: {fileID: 0}
+  cameraIntrinsicsPath: C:\Users\chriba\documents\Calibration\CameraIntrinsics_3.json
```
matthejo: >C:\Users\chriba\documents\Calibration\CameraIntrinsics_3.json [](start = 24, length = 61)

Probably don't want a path on your dev machine to be checked into the public repo

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodesManager.cs
chrisfromwork: we check the cancellation token before starting now

PR #124, opened 2019-07-17
Document calibration test window usage

chrisfromwork:
This review adds documentation to the calibration test window, it also has a few logic changes:

1) This review removes calibration status from the calibration test window. This information isn't when testing calibration and can be misleading to users.
2) Marker size is exposed as a field in the calibration test coded. Previously, the marker size was hard coded to 10cm which would make for less than ideal filming experiences when the marker used for localization was 10cm in length.


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemManager.cs
```diff
@@ -1,394 +1,396 @@
...
+
+            DebugLog($"Creating new SpatialCoordinateSystemParticipant, IPAddress: {endpoint.Address}, DebugLogging: {debugLogging}");
+
+            SpatialCoordinateSystemParticipant participant = new SpatialCoordinateSystemParticipant(endpoint, debugVisual, debugVisualScale);
+            participants[endpoint] = participant;
+            participant.ShowDebugVisuals = showDebugVisuals;
+            participant.SendSupportedLocalizersMessage(endpoint, localizers.Keys);
+
+            participant.SendSupportedLocalizersMessage(endpoint, localizers.Keys);
+
```
matthejo: Looks like this got duplicated as part of a merge, the call should only happen once

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CalibrationTestWindow.cs
```diff
@@ -35,7 +35,8 @@ internal class CalibrationTestWindow : CompositorWindowBase<CalibrationTestWindo
 
         private static readonly string holographicCameraIPAddressKey = $"{nameof(CalibrationTestWindow)}.{nameof(holographicCameraIPAddress)}";
         private static readonly string calibrationPlaybackIndexFilePathPreferenceKey = $"{nameof(CalibrationTestWindow)}.{nameof(indexFilePath)}";
-        private static readonly string calibrationPlaybackCalibrationFilePathPreferenceKey = $"{nameof(CalibrationTestWindow)}.{nameof(calibrationFilePath)}";
+        private static readonly string calibrationPlaybackCalibrationFilePathPreferenceKey = $"{nameof(CalibrationTestWindow)}.{nameof(calibrationFilePath)}";
+        private static readonly string markerSizeForPlaybackPreferenceKey = $"{nameof(CalibrationTestWindow)}.MarkerSizeForPlayback";
```
matthejo: >MarkerSizeForPlayback [](start = 110, length = 21)

Could make this '{nameof(markerSizeForPlayback)}' to match the pattern the other fields use

PR #126, opened 2019-07-17
Fix usage of Marshal.GetObjectForIUnknown on .NET Native builds

matthejo:
On .NET Native builds (the "master" build configuration of .NET Backed projects in Unity), using Marshal.GetObjectForIUnknown does not marshal any IInspectable object. This change works around this issue by specifically marshalling SpatialCoordinateSystem pointers through a native function. When marshalled as a specific IInspectable type, the marshaller will correctly wrap the IInspectable in the correct managed object.


src/SpectatorView.Unity/Assets/PhotoCapture/Scripts/WinRTExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// Provides extension methods that help marshal WinRT IInspectable objects that have been passed to managed code as an IntPtr.
+    /// On .NET Native, IInspectable pointers cannot be marshalled from native to managed code using Marshal.GetObjectForIUnknown.
+    /// This class calls into a native method that specifically marshals the type as a specific WinRT interface, which
+    /// is supported by the marshaller on both .NET Core and .NET Native.
+    /// </summary>
+    public static class WinRTExtensions
+    {
+        [DllImport("SpectatorView.WinRTExtensions.dll")]
+        public static extern void GetSpatialCoordinateSystem(IntPtr nativePtr, out SpatialCoordinateSystem coordinateSystem);
+    }
```
chrisfromwork: should we wrap this function and attempt the c# equivalent logic if this dll isn't found?

chrisfromwork:
We need to update the SpectatorView.Native readme to include steps on how to build this dll. Considering right now that we are trying to flush out documentation, it may be worth blocking on this until we have the associated documentation needed for this new dll/project.

chrisfromwork:
This looks great, one nit critique is that we may want to call out that building this dll is a good idea for all scenarios in our setup document. Regardless, this looks great!

PR #127, opened 2019-07-17
Enable specifying an ip address at runtime

chrisfromwork:
This review introduces the ability to specify a HoloLens ip address on mobile devices at runtime. It is a similar pattern to our recording service visuals. The Spectator view script now contains a default prefab serialized field that can generated a INetworkConfigurationVisual as needed. This visual can also be overridden in a new NetworkConfigurationSettings singleton that is added to the SpectatorViewSettings prefab.

1) An INetworkConfigurationVisual interface has been defined. Said interface can be used to setup a network connection with a custom ip address.
2) A MobileNetworkConfigurationVisual class and prefab have been created. This allows custom prefabs to be declared int he spectator view script.
3) A NetworkConfigurationSettings singleton has been created that allows specifying override behavior for custom INetworkConfigurationVisuals and whether to enable a network visual for mobile experiences.
4) A documentation file with a stale "Coming soon..." section has been fixed.

This has been tested on android and iOS


chrisfromwork:
ip address ui

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpectatorView.cs
```diff
@@ -72,7 +83,11 @@ public class SpectatorView : MonoBehaviour
         /// </summary>
         [Tooltip("Default prefab for creating a mobile recording service visual.")]
         [SerializeField]
-        public GameObject defaultMobileRecordingServiceVisualPrefab = null;
+#if UNITY_IOS || UNITY_ANDROID
+        private GameObject defaultMobileRecordingServiceVisualPrefab = null;
+#else
+        private GameObject defaultMobileRecordingServiceVisualPrefab;
+#endif
```
matthejo: Why is this difference needed? Can't we always have the version with the " = null;" at the end?

chrisfromwork: When you switch between ios/android and uwp it generates warnings based on whether there is a value declared thats unused or a value not declared. We could disable this warning which may look better than these if defs.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/UI/MobileNetworkConfigurationVisual.cs
```diff
@@ -0,0 +1,68 @@
...
+        public event NetworkConfigurationUpdatedHandler NetworkConfigurationUpdated;
+
+        private void OnEnable()
+        {
+            if (connectButton != null)
+            {
+                connectButton.onClick.AddListener(OnConnectButtonClick);
+            }
+
+            ipAddressField.text = ipAddress;
```
matthejo: >ipAddressField.text = ipAddress; [](start = 12, length = 32)

It would be nice to use PlayerPrefs to store and remember the previously-entered IP address so that you don't have to type it every time

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/StateSynchronizationMenuItems.cs
```diff
@@ -111,6 +111,7 @@ private static void EditCustomSettingsProperties()
                 hierarchyPrefab.AddComponent<BroadcasterSettings>();
                 hierarchyPrefab.AddComponent<SpatialLocalizationInitializationSettings>();
                 hierarchyPrefab.AddComponent<MobileRecordingSettings>();
+                hierarchyPrefab.AddComponent<NetworkConfigurationSettings>();
```
matthejo: >hierarchyPrefab.AddComponent<NetworkConfigurationSettings>(); [](start = 16, length = 61)

We might want to ensure these scripts are all present on this prefab, even if the prefab already existed. That way, as we add new functionality to SpectatorView, people who have projects already will get their prefabs updated to include the right scripts.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/UI/MobileNetworkConfigurationVisual.cs
```diff
@@ -0,0 +1,68 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Net;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace Microsoft.MixedReality.SpectatorView
+{
+    public class MobileNetworkConfigurationVisual : MonoBehaviour,
```
matthejo: >MobileNetworkConfigurationVisual [](start = 17, length = 32)

For the visual appearance of the default visual, I think it would be good to ensure there's an opaque or nearly-opaque background behind all of the text. This way the text is accessible and legible regardless of the background video you're holding the phone in front of.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpectatorView.cs
```diff
@@ -244,6 +270,59 @@ private bool TryCreateRecordingService(out IRecordingService recordingService)
 #endif
         }
 
+        private void SetupNetworkConfigurationVisual()
```
matthejo: >SetupNetworkConfigurationVisual [](start = 21, length = 31)

Do you think the network connection visual should re-show when a disconnect happens? Should there be additional UI or a public method that an app can call to re-show the connection UI?

chrisfromwork: This seems like a cool improvement. Maybe we could address this in a separate review? Users may also want to dynamically change the device that they are spectating as well.

chrisfromwork:
Notes: We should hold off on merging this PR until we have documentation on customizing UI.

chrisfromwork:
spectatorview

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpectatorView.cs
chrisfromwork: opened https://github.com/microsoft/MixedReality-SpectatorView/issues/133 for tracking

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpectatorView.cs
chrisfromwork: i'll leave this this way for now unless we want to disable this specific warning.

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/StateSynchronizationMenuItems.cs
chrisfromwork: This should now be updated

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpectatorView.cs
```diff
@@ -244,6 +270,59 @@ private bool TryCreateRecordingService(out IRecordingService recordingService)
...
+            }
+
+            mobileNetworkConfigurationVisual = Instantiate(mobileNetworkConfigurationVisualPrefab);
+            networkConfigurationVisual = mobileNetworkConfigurationVisual.GetComponentInChildren<INetworkConfigurationVisual>(true);
+            if (networkConfigurationVisual == null)
+            {
+                Debug.LogError("Network configuration visual was not found. No connection will be established.");
+                return;
+            }
+
```
andreiborodin: What happens to the created prefab at this point?

Could we have a different approach, where we instantiate if it's null, then show/hide?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/UI/INetworkConfigurationVisual.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.SpectatorView
+{
+    public delegate void NetworkConfigurationUpdatedHandler(object sender, string ipAddress);
```
andreiborodin: Can we use action here? And also, can we specify the type for sender instead of object?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/UI/NetworkConfigurationSettings.cs
```diff
@@ -0,0 +1,28 @@
...
+        private bool enableMobileNetworkConfigurationVisual = true;
+
+        /// <summary>
+        /// Prefab for creating a mobile network configuration visual.
+        /// </summary>
+        [Tooltip("Prefab for creating amobile network configuration visual, which replaces the defaultMobileNetworkConfigurationVisualPrefab on the SpectatorView component if set.")]
+        [SerializeField]
+        private GameObject overrideMobileNetworkConfigurationVisualPrefab = null;
+
+        public bool EnableMobileNetworkConfigurationVisual => enableMobileNetworkConfigurationVisual;
```
andreiborodin: We should really put the comments on the public properties instead of private fields. Generally we don't comment privates, and only add tooltip to private serializefields

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpectatorView.cs
chrisfromwork: hm, we probably need to destroy this visual at some point. But, having the ability to show/hide may prove beneficial on disconnect.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpectatorView.cs
```diff
@@ -244,6 +277,62 @@ private bool TryCreateRecordingService(out IRecordingService recordingService)
...
+        private void SetupNetworkConfigurationVisual()
+        {
+#if UNITY_ANDROID || UNITY_IOS
+            GameObject mobileNetworkConfigurationVisualPrefab = defaultMobileNetworkConfigurationVisualPrefab;
+            if (NetworkConfigurationSettings.IsInitialized && NetworkConfigurationSettings.Instance.OverrideMobileNetworkConfigurationVisualPrefab != null)
+            {
+                mobileNetworkConfigurationVisualPrefab = NetworkConfigurationSettings.Instance.OverrideMobileNetworkConfigurationVisualPrefab;
+            }
+
+            mobileNetworkConfigurationVisual = Instantiate(mobileNetworkConfigurationVisualPrefab);
```
andreiborodin: Instantiate only if null, right?

PR #132, opened 2019-07-18
Fix QR Localizers to not declare themselves as available on HoloLens 1

chrisfromwork:
There are a few changes in this review:

1) QR Code based localizers now don't declare themselves as supported on HoloLens 1.
2) The compositor UI has been updated to handle scenarios where no localizers are supported by a connected device.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindowBase.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindowBase.cs
@@ -204,7 +204,7 @@ namespace Microsoft.MixedReality.SpectatorView.Editor
             {
                 selectedLocalizerIndex = EditorGUILayout.Popup(selectedLocalizerIndex, localizerNames);
                 if (localizers != null &&
-                    localizers.Length> selectedLocalizerIndex)
+                    selectedLocalizerIndex < localizers.Length)
                 {
                     selectedLocalizerId = localizers[selectedLocalizerIndex].SpatialLocalizerId;
                     SpatialLocalizationSettingsGUI(deviceTypeLabel, selectedLocalizerIndex, localizers);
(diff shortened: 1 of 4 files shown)


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/MarkerVisual/QRCodeMarkerVisualDetectorSpatialLocalizer.cs
```diff
@@ -16,8 +16,8 @@ protected override bool IsSupported
         {
             get
             {
-#if QRCODESTRACKER_BINARY_AVAILABLE && UNITY_WSA
-                return true;
+#if QRCODESTRACKER_BINARY_AVAILABLE && UNITY_WSA && !UNITY_EDITOR
+                return (Windows.ApplicationModel.Package.Current.Id.Architecture != Windows.System.ProcessorArchitecture.X86); // HoloLens 1 is not supported.
```
matthejo: >// HoloLens 1 is not supported. [](start = 127, length = 31)

Do you think this should check specifically for ARM instead of for !x86?

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindowBase.cs
```diff
@@ -195,12 +203,12 @@ private void SpatialLocalizationGUI(string deviceTypeLabel, SpatialCoordinateSys
             GUILayout.BeginHorizontal();
             {
                 selectedLocalizerIndex = EditorGUILayout.Popup(selectedLocalizerIndex, localizerNames);
-                if (localizers != null)
+                if (localizers != null &&
+                    localizers.Length> selectedLocalizerIndex)
```
matthejo: >localizers.Length> selectedLocalizerIndex [](start = 20, length = 41)

Nit: I know it's logically equivalent but I would usually think of this as "selectedLocalizerIndex < localizers.Length"

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/MarkerVisual/QRCodeMarkerVisualDetectorSpatialLocalizer.cs
chrisfromwork: hm, technically this may one day be available for x64 desktop scenarios. I think for now this will suffice.

andreiborodin: Smart!

---
In reply to: [305533927](https://github.com/microsoft/MixedReality-SpectatorView/pull/132#discussion_r305533927) [](ancestors = 305533927)

PR #141, opened 2019-07-24
Update the readme setup steps to have a correct GoogleARCore location

wiwei:
Walking through these steps, I hit an issue where GoogleARCore wasnt a valid folder to use - GoogleARCore doesn't exist as a folder at that location based on the setup steps in the tooling.

I'm not 100% sure if we should only be including GoogleARCore, or if we want to include the rest of the ARCore assets


wiwei:
I think similar things will need to happen for the scripts as well, though I'll probably wait to hear back what the intended target was here (i.e. just GoogleARCore, or the rest of the ARCore package)

chrisfromwork:
FWIW, it's bad practice to make your own branches in this repo. Although you have permissions to do so, you should create your own fork of this repo.

Issue #99, opened 2019-07-08
SpectatorViewOpenCVInterface shouldn't attempt to access the SpectatorView.OpenCV dll unless the ArUco marker detector spatial localizer is getting used.


chrisfromwork:

chrisfromwork:
There are also issues where the attempt at accessing a capability for QR Codes doesn't work with HoloLens 1. All marker detector localizers should only enable marker detectors once they have been called to start localizing.

PR #148, opened 2019-07-25
Update localizer prefab to have disabled marker detectors

chrisfromwork:
We have some issues in our spectator view localizer prefab where marker detectors initialize on platforms that they are not supported. To fix this, marker detector monobehaviours are now set to disabled by default. When marker detection starts, the marker detectors enable themselves. This review contains the following:

1) Marker detectors now enable/disable themselves based on start/stop calls
2) The ArUco marker detector is updated to have more robust asynchronous start/stop when initializing a HololensCamera.

This pull request addresses issue: https://github.com/microsoft/MixedReality-SpectatorView/issues/99


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/ArUco/HoloLens/ArUcoMarkerDetector.cs
```diff
@@ -84,13 +88,10 @@ public MarkerPositionBehavior MarkerPositionBehavior
             }
         }
 
-        private void Awake()
+        protected void OnEnable()
```
matthejo: >protected [](start = 8, length = 9)

I think these could all be consistently private. If it was expected that someone might override this class, all of the MonoBehaviour methods should be "protected virtual"

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/ArUco/HoloLens/ArUcoMarkerDetector.cs
```diff
@@ -131,13 +137,15 @@ private void DebugLog(string message)
...
+            enabled = true;
+
 #if UNITY_WSA
             if (!_detecting)
             {
                 _detecting = true;
                 _markerObservations.Clear();
                 DebugLog("Starting ArUco marker detection");
-                SetupCamera();
+                SetupCameraAsync();
```
matthejo: >SetupCameraAsync(); [](start = 16, length = 19)

I believe this should "await SetupCameraAsync()". If a method returns void and an exception is thrown, it will result in an unhandled exception error, but if an async method returns a Task and that method throws an exception, that Task will become Faulted but the exception won't be rethrown unless someone awaits the task.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/ArUco/HoloLens/ArUcoMarkerDetector.cs
```diff
@@ -147,13 +155,18 @@ public void StartDetecting()
...
         {
-            _detecting = false;
+            if (_detecting)
+            {
+                _detecting = false;
 #if UNITY_WSA
-            DebugLog("Stopping ArUco marker detection");
-            CleanUpCamera();
+                DebugLog("Stopping ArUco marker detection");
+                CleanUpCameraAsync();
```
matthejo: >CleanUpCameraAsync(); [](start = 16, length = 21)

This should "await CleanUpCameraAsync()"

PR #149, opened 2019-07-26
Add note to ASA instructions that applications should use access/authentication tokens

chrisfromwork:
This is related to https://github.com/microsoft/MixedReality-SpectatorView/issues/131

### DIFF
--- a/doc/SpectatorView.Setup.md
+++ b/doc/SpectatorView.Setup.md
@@ -64,7 +64,7 @@ Not all spatial alignment strategies support all platforms. See the chart below 
 
 ![Marker](images/ASAInspector.png)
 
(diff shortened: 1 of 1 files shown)


doc/SpectatorView.Setup.md
```diff
@@ -64,6 +64,8 @@ Not all spatial alignment strategies support all platforms. See the chart below
 
 ![Marker](images/ASAInspector.png)
 
+>Note: Use of an Account Id and Account Key can accelerate your development process. However, hardcoding these values into your application isn't a safe practice and should be avoided for enterprise deployed solutions. For your end application, its suggested to use an Access or Authentication token. More information on how to setup and use AAD tokens with ASA can be found [here](https://docs.microsoft.com/en-us/azure/spatial-anchors/concepts/authentication?tabs=csharp#azure-ad-user-authentication).
```
wiwei:
```suggestion
> Note: Use of an Account Id and Account Key can accelerate your development process. However, hardcoding these values into your application isn't a safe practice and should be avoided for enterprise deployed solutions. For your end application, its suggested to use an Access or Authentication token. More information on how to setup and use AAD tokens with ASA can be found [here](https://docs.microsoft.com/en-us/azure/spatial-anchors/concepts/authentication?tabs=csharp#azure-ad-user-authentication).
```

Issue #139, opened 2019-07-23
Suggesting symbolic linked directories does not work for multi-user teams


chrisfromwork:
We need to add instructions for submoduling the MixedReality-SpectatorView repo. Setting up items with symbolic linked directories doesn't make for a good multi-dev team experience.

Issue #136, opened 2019-07-23
Documentation issues from hackathon 7.23.2019


chrisfromwork:
~~It may not be immediately obvious how to determine what branch has been cloned/checked out when going through repo setup steps. Add instructions calling out "git branch" to see what branch a user is in.~~ https://github.com/microsoft/MixedReality-SpectatorView/pull/150

~~It's not obvious how to configure spectator view for a multi-scene experience.~~ Tracked with https://github.com/microsoft/MixedReality-SpectatorView/issues/151

~~The powershell window disappearing after running the setup scripts can be misconstrued as a failure.~~ Fixed in #150 

~~Powershell may require running "Set-ExecutionPolicy Unrestricted" to enable setup scripts.~~ This was related to the bat file being configured incorrectly. Fixed in #150 

It's unclear that most dependencies were obtained through the SetupRepository script.

It's not immediately obvious how to locate unity player settings/build settings. We should have steps calling out each individual player settings platform to add preprocessor directives too.

We need to call out that you need android/ios build tools installed. We could add pictures showing what the inspector looks like for player settings when these aren't installed.

It's unclear how to organize scene content for a GameObjectHierarchyBroadcaster, We should add instructions around synchronizing all content.

PR #150, opened 2019-07-26
Fix documentation to instruct consumers to setup submodules

chrisfromwork:
This review contains a few changes:

1) AddDependencies.ps1 is updated to use relative paths in order to play well with checking in submodule based symbolic links
2) SetupRepository.bat is edited to run in the command window. It also now requires user input to end its execution. User feedback stated that it was unclear whether the powershell script succeeded/failed when it ran in a separate window that immediately closed after completion.
3) Documentation is updated to instruct users to setup submodules. These set up steps now include images.
4) Symbolic link related helper functions have been moved to SymbolicLinkHelpers.ps1 so that they could be reused for fixing symbolic links in a consumer's project repository.

This pull request addresses https://github.com/microsoft/MixedReality-SpectatorView/issues/139, https://github.com/microsoft/MixedReality-SpectatorView/issues/136


tools/Scripts/FixSymbolicLinks.ps1
```diff
@@ -0,0 +1,22 @@
...
+dir Get-Location -Recurse -File | ?{$_.LinkType -eq "SymbolicLink" } | Restore-SymbolicLink
+
+# If any links were created before the repository was configured to use
+# symlinks, those links need to be restored.
+git status --porcelain | Restore-SymbolicLinkFromTypeChange
+
+Write-Host "`n"
+Write-Host -NoNewLine 'Symbolic links fixed. Press any key to continue...';
+$null = $Host.UI.RawUI.ReadKey('NoEcho,IncludeKeyDown');
+Write-Host "`n"
```
matthejo: Since this is the same logic as exists in SetupRepository.ps1, should all of this logic just be contained in SymbolicLinkHelpers and then this PS script can call a single method in SymbolicLinkHelpers?

README.md
```diff
@@ -35,59 +43,74 @@ The MixedReality-SpectatorView repository uses Unity packages, git submodules an
...
-## Setting up your own project
-
 ### Adding references to your own project
 
 After obtaining a local clone of the MixedReality-SpectatorView repository and resolving its external dependencies (see above), the suggested mechanism for referencing the code is by adding symbolic linked directories to your Unity project's Assets folder. You can do this with the following:
 
+> Note: Symbolic linked directories should be setup as relative paths. Using relative paths should allow directories to resolve correctly regardless of where you or your team members clone your project repository in their local file systems. The instructions below demonstrate setting up symbolic linked directories based on the following paths:
+>* **Project repository directory:** c:\Your\Unity\Project
+>* **Project Assets directory:** c:\Your\Unity\Project\Assets
+>* **MixedReality-SpectatorView submodule directory:** c:\Your\Unity\Project\sv
```
matthejo: >sv [](start = 77, length = 2)

Are we recommending "sv" as a name because of path limits?

chrisfromwork: Yeah, that's the main reason.

README.md
```diff
@@ -35,59 +43,74 @@ The MixedReality-SpectatorView repository uses Unity packages, git submodules an
...
-* `mklink /D "MixedReality-QRCodePlugin" "%PathToRepo%\MixedReality-SpectatorView\external\MixedReality-QRCodePlugin"`
-* `mkdir AzureSpatialAnchors.Resources`
-* `cd AzureSpatialAnchors.Resources`
-* `mklink /D "android-logos" "%PathToRepo%\MixedReality-SpectatorView\external\Azure-Spatial-Anchors-Samples\Unity\Assets\android-logos"`
-* `mklink /D "logos" "%PathToRepo%\MixedReality-SpectatorView\external\Azure-Spatial-Anchors-Samples\Unity\Assets\logos"`
+4. `cd c:\Your\Unity\Project\Assets`
+5. `mklink /D "MixedReality-SpectatorView" "..\sv\src\SpectatorView.Unity\Assets"`
+6. `mklink /D "ARKit-Unity-Plugin" "..\sv\external\ARKit-Unity-Plugin"`
+7. `mklink /D "AzureSpatialAnchorsPlugin" "..\sv\external\Azure-Spatial-Anchors-Samples\Unity\Assets\AzureSpatialAnchorsPlugin"`
+8. `mklink /D "GoogleARCore" "..\sv\external\ARCore-Unity-SDK\Assets"`
```
wiwei: Nit: Can we also use the same folder name (i.e. ARCore-Unity-SDK) instead of GoogleARCore?

I see that there is a subfolder called GoogleARCore, and when I saw this during my attempt to get this setup, I wasn't sure if the intent was to take the entire folder, or just that subfolder

chrisfromwork: Good catch, we actually want to point to the GoogleARCore folder hence the name. The above line is just wrong.

PR #160, opened 2019-07-30
Add a ComponentBroadcasterService for RectMask2D

matthejo:
This change adds a broadcaster and observer for RectMask2D. This change also modifies the SpectatorView prefab to refer to the StateSynchronizationAssets prefab so that future new services can be added only in one prefab.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/RectMask2D/RectMask2DService.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/RectMask2D/RectMask2DService.cs
@@ -8,7 +8,7 @@ namespace Microsoft.MixedReality.SpectatorView
 {
     internal class RectMask2DService : ComponentBroadcasterService<RectMask2DService, RectMask2DObserver>
     {
-        public static readonly ShortID ID = new ShortID("RM2");
+        public static readonly ShortID ID = new ShortID("RMK");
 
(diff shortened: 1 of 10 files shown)


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/RectMask2D/RectMask2DService.cs
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace Microsoft.MixedReality.SpectatorView
+{
+    internal class RectMask2DService : ComponentBroadcasterService<RectMask2DService, RectMask2DObserver>
+    {
+        public static readonly ShortID ID = new ShortID("RMK");
```
chrisfromwork: >RMK [](start = 57, length = 3)

Do we have any mechanism for ensuring that no collisions occur in ids?

matthejo: At runtime you'll get an exception about a key already being added to a dictionary. I could make that a more-explicit error though

chrisfromwork: definitely not required in this review, just was curious :)

PR #172, opened 2019-08-02
Re-enable the normal rendering path when not using quad view

matthejo:
The way that the "quad" view rendering works affects how transparency is handled for partially-transparent holograms. Previously, the rendering would:

1. Fill the camera's render texture with the real-world video
2. Unity renders holograms onto the real-world video
3. Re-draw the real-world video onto the resulting texture to create global hologram opacity

Because in this path the holograms are alpha-blended with the real-world background, their transparencies appear correct. This is what these intermediate textures look like.
image

The quadrant view approach works differently:
1. The camera is filled with black
2. Unity renders holograms onto the black texture
3. The holograms are blended with the real-world video using the alpha values from the resulting texture, including global hologram opacity

Because at step 2 the holograms are blended with black, they end up overall darker than if they'd been initially blended against the real-world video.  Here's a comparison in the same case as the previous image.

image

This PR restores using the original rendering path when not using quad video recording output. This allows for the majority of cases to have the more-correct alpha blending with the real world instead of with black. For the quadrant recording case, it is up to the end application to correctly handle alpha values of any partially-transparent holograms.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
@@ -118,9 +118,13 @@ namespace Microsoft.MixedReality.SpectatorView.Editor
 
                 EditorGUILayout.BeginHorizontal("Box");
                 {
-                    string[] compositionOptions = new string[] { "Final composited texture", "Quadrant with intermediate textures" };
-                    GUIContent renderingModeLabel = new GUIContent("Display texture", "Choose between displaying the composited video texture or seeing intermediate textures displayed in 4 sections (bottom left: input video, top left: opaque hologram, top right: hologram alpha mask, bottom right: hologram alpha-blended onto video)");
+                    string[] compositionOptions = new string[] { "Final composite", "Intermediate textures" };
+                    GUIContent renderingModeLabel = new GUIContent("Preview display", "Choose between displaying the composited video texture or seeing intermediate textures displayed in 4 sections (bottom left: input video, top left: opaque hologram, top right: hologram alpha mask, bottom right: hologram alpha-blended onto video)");
                     textureRenderMode = EditorGUILayout.Popup(renderingModeLabel, textureRenderMode, compositionOptions);
+                    if (compositionManager != null && compositionManager.TextureManager != null)
+                    {
+                        compositionManager.TextureManager.IsQuadrantVideoOutputRequired = textureRenderMode == (int)VideoRecordingFrameLayout.Quad;
+                    }
                     FullScreenCompositorWindow fullscreenWindow = FullScreenCompositorWindow.TryGetWindow();
                     if (fullscreenWindow != null)
                     {
@@ -155,8 +159,8 @@ namespace Microsoft.MixedReality.SpectatorView.Editor
                     {
                         bool wasEnabled = GUI.enabled;
                         GUI.enabled = compositionManager != null && !compositionManager.IsRecording();
-                        string[] compositionOptions = new string[] { "Final composited texture", "Quadrant with intermediate textures" };
-                        GUIContent renderingModeLabel = new GUIContent("Video recording texture", "Choose between displaying the composited video texture or seeing intermediate textures displayed in 4 sections (bottom left: input video, top left: opaque hologram, top right: hologram alpha mask, bottom right: hologram alpha-blended onto video)");
+                        string[] compositionOptions = new string[] { "Normal", "Split channels" };
+                        GUIContent renderingModeLabel = new GUIContent("Video output mode", "Choose between recording the composited video texture or recording intermediate textures displayed in 4 sections (bottom left: input video, top left: opaque hologram, top right: hologram alpha mask, bottom right: hologram alpha-blended onto video)");
                         compositionManager.VideoRecordingLayout = (VideoRecordingFrameLayout)EditorGUILayout.Popup(renderingModeLabel, (int)compositionManager.VideoRecordingLayout, compositionOptions);
                         GUI.enabled = wasEnabled;
                     }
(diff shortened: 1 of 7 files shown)


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/TextureManager.cs
```diff
@@ -64,6 +71,14 @@ public class TextureManager : MonoBehaviour
         /// </summary>
         private RenderTexture displayOutputTexture = null;
 
+        /// <summary>
+        /// Gets whether or not holograms should be rendered on a black background (which allows for post-processing in quadrant mode)
+        /// or against the video background (which allows for correct alpha blending for partially-transparent holograms).
+        /// </summary>
+        private bool IsVideoOutputQuadrantMode => Compositor != null && Compositor.VideoRecordingLayout == VideoRecordingFrameLayout.Quad;
```
chrisfromwork: >IsVideoOutputQuadrantMode [](start = 21, length = 25)

This flag name is extremely confusing. From what I can tell, this flag is an indication of whether video recording is in quadrant mode compared to whether or not the capture card will output quadrant mode. Can we change this to IsVideoRecordingQuadrantMode?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/TextureManager.cs
```diff
@@ -19,6 +19,13 @@ public class TextureManager : MonoBehaviour
 #if UNITY_EDITOR
         public CompositionManager Compositor { get; set; }
 
+        /// <summary>
+        /// Gets or sets whether or not the quadrant video output is required.
+        /// The quadrant frame will not be rendered when this is not set and when
+        /// the video recording mode does not require it.
+        /// </summary>
+        public bool IsQuadrantVideoOutputRequired { get; set; }
```
chrisfromwork: >IsQuadrantVideoOutputRequired [](start = 20, length = 29)

to remain consistent with the flag below, can we call this IsVideoOutputQuadrantMode?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/TextureManager.cs
```diff
@@ -249,9 +266,18 @@ private void OnPreRender()
...
             Graphics.Blit(CurrentColorTexture, colorRGBTexture, CurrentColorMaterial);
 
-            // Clear the camera's background by blitting using a shader that simply
-            // clears the target texture without reading from the source texture.
-            Graphics.Blit(null, spectatorViewCamera.targetTexture, textureClearMat);
+            if (IsVideoOutputQuadrantMode)
+            {
+                // Clear the camera's background by blitting using a shader that simply
+                // clears the target texture without reading from the source texture.
+                Graphics.Blit(null, spectatorViewCamera.targetTexture, textureClearMat);
```
chrisfromwork: nit: maybe call out that this is only required for recording in quad compared to displaying in quad?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/TextureManager.cs
```diff
@@ -273,22 +299,36 @@ private void OnPostRender()
             // force set this every frame as it sometimes get unset somehow when alt-tabbing
             renderTexture = sourceTexture;
 
-            //composite hologram onto video
-            BlitCompositeTexture(renderTexture, colorRGBTexture, compositeTexture);
+            if (IsVideoOutputQuadrantMode)
+            {
+                // composite hologram onto video
+                BlitCompositeTexture(renderTexture, colorRGBTexture, compositeTexture);
```
chrisfromwork: again, it seems worth calling out that this is only required for recording.

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
```diff
@@ -118,9 +118,13 @@ private void CompositeGUI()
...
+                    string[] compositionOptions = new string[] { "Final composite", "Intermediate textures" };
+                    GUIContent renderingModeLabel = new GUIContent("Preview display", "Choose between displaying the composited video texture or seeing intermediate textures displayed in 4 sections (bottom left: input video, top left: opaque hologram, top right: hologram alpha mask, bottom right: hologram alpha-blended onto video)");
                     textureRenderMode = EditorGUILayout.Popup(renderingModeLabel, textureRenderMode, compositionOptions);
+                    if (compositionManager != null && compositionManager.TextureManager != null)
+                    {
+                        compositionManager.TextureManager.IsQuadrantVideoOutputRequired = textureRenderMode == (int)VideoRecordingFrameLayout.Quad;
+                    }
```
chrisfromwork: should we consider another flag on the composition manager called VideoOutputLayout, similar to VideoRecordingLayout?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/TextureManager.cs
matthejo: I changed the name to "IsQuadrantVideoFrameNecessary". The quadrant frame might be necessary because it's required for video recording, or because it's being previewed on screen. The output is never "quadrant" from the capture card

---
In reply to: [310673196](https://github.com/microsoft/MixedReality-SpectatorView/pull/172#discussion_r310673196) [](ancestors = 310673196)

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
matthejo: Right now the video output layout is always "Composite" because the BlackMagic 4K capture card does not support outputting 4K unless the input video is 4K, and I don't think we want to output 1080p quadrant video (each sub-frame would be only 720p).

---
In reply to: [310675484](https://github.com/microsoft/MixedReality-SpectatorView/pull/172#discussion_r310675484) [](ancestors = 310675484)

PR #175, opened 2019-08-05
Allow overriding the composited camera feed to show other textures such as videos/webcam feeds.

chrisfromwork:
In some demo scenarios, users may want to specify a feed other than the camera + hologram feed to show to the audience. This review adds the ability to override the output texture in order for this to be possible.

1) An overrideOutputTexture is defined for the TextureManager. Users can now specify video textures, etc to output to the audience.
2) A new previewTexture is defined for the TextureManager. This allows the compositor editor gui to consistently grab the same texture that is output to the capture card when not in quadrant output mode.


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/TextureManager.cs
```diff
@@ -41,6 +41,11 @@ public class TextureManager : MonoBehaviour
         /// </summary>
         public RenderTexture compositeTexture { get; private set; }
 
+        /// <summary>
+        /// The texture output to a capture card when quadrant view isn't enabled
```
matthejo: >The texture output to a capture card when quadrant view isn't enabled [](start = 12, length = 69)

The output to the capture card is currently always the same, it doesn't matter if quadrant view is enabled or not

Issue #98, opened 2019-07-08
Remove unity development console from debug builds.


chrisfromwork:

chrisfromwork:
The main MRTK code base does not result in a development console.

chrisfromwork:
If someone does a first pass investigating this and determines that there is no quick fix, this may not land with the 8.15 milestone

chrisfromwork:
This is a .NET build problem related to debug and release. il2cpp & master .NET builds should not experience this.

chrisfromwork:
It may be worth disabling this in our samples: https://docs.unity3d.com/ScriptReference/Debug-developerConsoleVisible.html

PR #178, opened 2019-08-05
Add a flag for disabling the developer console

chrisfromwork:
This pr addresses https://github.com/microsoft/MixedReality-SpectatorView/issues/98

This review contains the following:
1) hideDevelopmentConsole is added to the SpectatorView.cs script. When checked, SpectatorView checks whether the development console is visible every update. If the console is visible, its hidden. The default value for this is unchecked.
2) hideDevelopmentConsole is checked in SpectatorView.HoloLens.unity and Finished_Scene.unity. This results in the development console being hidden in our sample projects.

Note: The development console is still shown during the unity splash screen if any errors are encountered prior to Starting/Updating the SpectatorView prefab.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpectatorView.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpectatorView.cs
@@ -118,9 +118,9 @@ namespace Microsoft.MixedReality.SpectatorView
         [SerializeField]
         private bool debugLogging = false;
 
-        [Tooltip("Check to hide the development console every update.")]
+        [Tooltip("Check to hide the developer console every update.")]
         [SerializeField]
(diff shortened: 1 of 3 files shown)


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpectatorView.cs
```diff
@@ -118,6 +118,10 @@ public class SpectatorView : MonoBehaviour
         [SerializeField]
         private bool debugLogging = false;
 
+        [Tooltip("Check to hide the developer console every update.")]
```
matthejo: >[Tooltip("Check to hide the developer console every update.")] [](start = 8, length = 62)

I think it might be worth clarifying that this "developer console" is the on-device console, not related to the console pane in Unity

Issue #186, opened 2019-08-15
Abstract SpectatorView network transport layer


stephenatwork:
Allow SpectatorView to use user-provided network transport instead of managing sockets directly.

PR #187, opened 2019-08-15
Extract network interfaces

stephenatwork:
This addresses #186 

TODO: Check if things still work - I only checked it built.

Deferred TODOs, not for this PR.
* Rename SocketEndpoint to ISocketEndpoint (didn't do here because of the massive churn).
* How to switch between IConnectionManager implementations? We don't need to consider this until we have another implementation. When we do, I note there are 2 [RequireComponent(typeof(TCPConnectionManager))] which will need replacing since RequireComponenent only works with instantiable types.

Breaking Change Details:

This change introduces a new parent interface of TCPConnectionManager, namely IConnectionManager. All public methods of TCPConnectionManager were hoisted into the base interface so this change should be transparent.

This change also moves the existing SocketEndpoint to a new class TCPSocketEndpoint and extracts the public members into a new interface SocketEndpoint. As before, all public methods are preserved except that SocketEndpoint is no longer constructible - one must replace all calls to ```new SocketEndpoint(...)``` with ```new TCPSocketEndpoint(...)```. The only callers in the codebase are within TCPConnectionManager.

More? I'd appreciate your help here @chrisfromwork , I'm not familiar enough with this codebase yet to know if there are other gotchas.


chrisfromwork:
Please provide Breaking Change Details. We need a note on why this change was introduced/what extension surface it impacts. We also need steps for migrating to this new codebase. https://github.com/microsoft/MixedReality-SpectatorView/pull/213 should serve as an example/please include these notes in the pull request description.

PR #212, opened 2019-09-05
Move build tools templates/scripts into the master branch to start having master builds

chrisfromwork:
This review contains the following logic:

1) Duplicated/slightly modified Unity *.cs scripts from the MRTK repo that enable building from the command line.
2) Powershell/bat scripts that allow building Unity projects as well as native plugins from the command line.
3) Yaml/yml files that can be used to setup azure pipeline jobs.
4) Native projects that depend on vcpkg generated opencv now have PropertyGroup items defined for their AdditionalIncludeDirectories, AdditionalLibraryDirectories and AdditionalDependencies. This allows specifying the vcpkg output as a parameter in the MSBuild step within buildnative.yml

For reading:
1) tools/ci/src/BuildTools.Unity/Assets contains the duplicated MRTK Unity logic.
2) tools/ci contains the default yaml files that are used by our azure pipelines.
3) tools/ci/templates contains the different build configurations specific to the Spectator View project.
4) tools/ci/scripts contains scripts that were related to ci. Some items like genericHelpers.ps1 may prove reusable with other projects. spectatorViewHelpers.ps1 is intended to contain build logic steps that are specific to the spectator view project.


tools/Scripts/SetupRepository.ps1
```diff
@@ -1,28 +1,27 @@
 param(
-    [switch] $iOS
+    [switch] $iOS,
+    [switch] $NoDownloads
 )
 
-. "$PSScriptRoot\SymbolicLinkHelpers.ps1"
-. "$PSScriptRoot\ExternalDependencyHelpers.ps1"
+. "$PSScriptRoot\SetupRepositoryFunc.ps1"
```
macborow: If the *Func.ps1 files contain only shared functions perhaps they should be renamed to *.psm1 and included with Import-Module?

tools/ci/scripts/buildAllLocal.ps1
```diff
@@ -0,0 +1,23 @@
+$MSBuild = "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\amd64\MSBuild.exe"
+$UnityPath = "C:\Program Files\Unity\Hub\Editor\2018.3.14f1\Editor\Unity.exe"
```
macborow: It's not great that those paths are hard coded. If we can't detect those, perhaps it might be a good idea to set up a shared config file somewhere? This is yet another place where Unity version needs to be updated if build requirements change.
TBH, if you don't want to do spend time on doing this now perhaps it's worth adding a TODO comment or creating a work item for this.

tools/ci/scripts/buildNativeProjectLocal.ps1
```diff
@@ -0,0 +1,85 @@
...
+}
+elseif (!$MSBuild)
+{
+    Write-Error Unable to locate MSBuild.exe
+    exit 1;
+}
+
+if (!(Get-Command "nuget"))
+{
+    Write-Error "NuGet.exe does not seem to be isntalleda as a command on this computer."
```
macborow: isntalleda - typo

tools/ci/scripts/genericHelpers.ps1
```diff
@@ -0,0 +1,297 @@
...
+
+Write-Output '====================================================='
+Write-Output '           Unity Build Player Finished               '
+Write-Output '====================================================='
+
+   # Changing this variable will break reporting scripts
+   $Succeeded.Value = ($proc.ExitCode -eq 0)
+}
+
+function ConfirmEditorCompiles
```
macborow: The only difference between this function and BuildUnityProject seems to be extra logging? Is there really a need to keep both of them?

tools/ci/templates/common.yml
```diff
@@ -0,0 +1,69 @@
...
+    ProjectPath: '${{ parameters.ProjectPath }}'
+    ToolsAssetsPath: '${{ parameters.ToolsAssetsPath }}'
+    Platform: 'UWP'
+    Arch: 'ARM'
+    ScriptingBackend: 'IL2CPP'
+    SceneList: '${{ parameters.ProjectPath }}\Assets\MixedReality-SpectatorView\SpectatorView\Scenes\SpectatorView.HoloLens.unity'
+    Define: ''
+    UnityArgs: ''
+
+# # Build UWP ARM IL2CPP ASA QR
```
macborow: Any reason for leaving this commented out instead of removing?

tools/ci/scripts/buildAllLocal.ps1
chrisfromwork: when i set up some of the buildNativeLocal.ps1 logic I had these as fallback values if no value was specified. Long term it seems like we could use some of the searching functionality that exists for unity in the pipeline build to find these executables. I'll take another pass on this before checking in.

tools/ci/templates/common.yml
chrisfromwork: ah, i want to uncomment these additional passes once one succeeds on an agent. Good catch/I should probably enable these now.

FWIW - these builds are currently failing because our agents dont support symbolic linked directories. were working to fix this now.

PR #213, opened 2019-09-05
Add tests for video camera video recording and fix subsequent recording

chrisfromwork:
This review contains a few items:

1) SpectatorView.Compositor.UnityPlugin no longer tries to interpolate between video and audio times. There is now an assumption that audio samples/video samples will be provided by the unity editor based on the same start time.

2) Audio sample time provided from the Unity editor is now based off of the captured frame index + frame sample lengths. This prevents needing to interpolate between the AudioSettings.dspTime and the CaptureFrameIndex time. Previous calculations for doing this in the native plugin did not work correctly and seemed unnecessary. Interpolations between AudioSettings.dspTime and CaptureFrameIndex time can be conducted in the Unity editor as needed.

3) We now have tests to ensure that textures get created for both the elgato and black magic design frame providers when starting the SpectatorViewCompositor scene.

4) We now have tests for recording and playing videos recorded with the SpectatorViewCompositor scene.

5) SpectatorView.Compositor & SpectatorView.Compositor.UnityPlugin now return video file names when starting recording so that tests can cleanup generated files.

### Notes:
This change fixes video encoding issues that occur when recording videos with the Video Camera filming solution. It redefines the UnityCompositorInterface and is therefore a breaking change (requires recompiling associated dlls). This break allows the editor to know the name of the video file recorded, which allows for playmode test cleanup.

### Migration Instructions:
To migrate to codebases using this change, users need to do the following:
1) x64 Release versions of SpectatorView.Compositor.dll and SpectatorView.Compositor.UnityPlugin.dll.
2) Close any instances of Unity
3) Run tools\Scripts\CopyPluginsToUnity.bat (this will update the dlls within SpectatorView.Native within your local SpectatorView.


src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
```diff
@@ -392,8 +444,9 @@ void VideoEncoder::StopRecording()
     {
         while (!videoQueue.empty())
         {
-            videoQueue.pop();
+			videoQueue.pop();
         }
+		OutputDebugString(L"Cleared video queue\n");
```
matthejo: >OutputDebugString(L"Cleared video queue\n"); [](start = 2, length = 44)

Should this be in an #if _DEBUG block as well?

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/CompositorInterface.cpp
```diff
@@ -176,10 +176,14 @@ void CompositorInterface::StartRecording(VideoRecordingFrameLayout frameLayout)
     }
 
     videoIndex++;
-    videoRecordingStartTime = compositeFrameIndex * GetColorDuration();
     audioRecordingStartTime = -1.0;
 
     std::wstring videoPath = DirectoryHelper::FindUniqueFileName(outputPath, L"Video", L".mp4", videoIndex);
+	if (videoPath.size() <= lpFilePathLength)
+	{
+		memcpy(lpFilePath, videoPath.data(), lpFilePathLength);
```
matthejo: >memcpy(lpFilePath, videoPath.data(), lpFilePathLength); [](start = 2, length = 55)

Can we use memcpy_s here?

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/CompositorInterface.cpp
```diff
@@ -159,7 +159,7 @@ bool CompositorInterface::InitializeVideoEncoder(ID3D11Device* device)
     return videoEncoder1080p->Initialize(device) && videoEncoder4K->Initialize(device);
 }
 
-void CompositorInterface::StartRecording(VideoRecordingFrameLayout frameLayout)
+void CompositorInterface::StartRecording(VideoRecordingFrameLayout frameLayout, LPWSTR lpFilePath, int lpFilePathLength)
```
matthejo: I know this would be a larger change, but I'd prefer a design where the caller to StartRecording can specify the target path. This would mean that the logic in FindUniqueFileName would need to be written in C# in the compositor instead of here, but it would make this library more flexible in that you could choose to record wherever you want without rebuilding the native compositor.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/UnityCompositorInterface.cs
```diff
@@ -71,8 +71,8 @@ internal static class UnityCompositorInterface
         [DllImport(CompositorPluginDll, CharSet = CharSet.Unicode)]
         public static extern void TakeRawPicture(string path);
 
-        [DllImport(CompositorPluginDll)]
-        public static extern void StartRecording(int frameLayout);
+        [DllImport(CompositorPluginDll, CharSet = CharSet.Unicode)]
+        public static extern void StartRecording(int frameLayout, string filePath, int filePathLength);
```
matthejo: >public static extern void StartRecording(int frameLayout, string filePath, int filePathLength); [](start = 8, length = 95)

I think the recommended way to marshal an "out" string is to use a StringBuilder and not a string. String is meant to be immutable in C#, and writing directly to its buffer from native code may cause unexpected behavior. StringBuilder is meant to be a writable string buffer.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/CompositionManager.cs
```diff
@@ -612,10 +612,12 @@ public bool IsRecording()
             return UnityCompositorInterface.IsRecording();
         }
 
-        public void StartRecording()
+        public string StartRecording()
         {
             TextureManager.InitializeVideoRecordingTextures();
-            UnityCompositorInterface.StartRecording((int)VideoRecordingLayout);
+            string filePath = new string(' ', 1024);
+            UnityCompositorInterface.StartRecording((int)VideoRecordingLayout, filePath, filePath.Length);
```
matthejo: If you change the marshalling to use StringBuilder, you'd just create a new StringBuilder(1024) and then pass filePath.Capacity for the length parameter

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/CompositionManager.cs
```diff
@@ -638,6 +640,13 @@ private void StopRecordingAudio()
                 audioMemoryStream = null;
             }
         }
+        
+        private string ReduceFilePath(string filePath)
+        {
+            const string mp4Ext = ".mp4";
+            filePath = filePath.Trim().Replace("\\\\", "\\");
+            return filePath.Substring(0, filePath.LastIndexOf(mp4Ext) + mp4Ext.Length);
```
matthejo: I'm confused about why this is necessary:
1. When are there double "\\" instances in the path? Is that indicative of a bug in the native code that builds the path?
2. Why is there a need to trim a substring that cuts something after the ".mp4" in the path?

tools/Scripts/CopyPluginsToUnity.ps1
```diff
@@ -1,3 +1,7 @@
+param(
+    [switch] $DebugCompositor
```
matthejo: >[switch] $DebugCompositor [](start = 4, length = 25)

Maybe this should just be a "BuildFlavor" and it can be either Release or Debug, and all of the components can use the same type? Otherwise, it seems like it will be weird if we want to add a "DebugOpenCV" flag later

matthejo:
> 

Remove this if there's no tests yet?

---
Refers to: src/SpectatorView.Unity/Assets/SpectatorView.Tests/EditorIntrinsicsCalibrationTests.cs:19 in 7aa64a9. [](commit_id = 7aa64a95d8b71297b8d66472ba31d7dea98393ff, deletion_comment = False)

matthejo:
> 

Remove this if there's no tests yet?

---
Refers to: src/SpectatorView.Unity/Assets/SpectatorView.Tests/EditorExtrinsicsCalibrationTests.cs:19 in 7aa64a9. [](commit_id = 7aa64a95d8b71297b8d66472ba31d7dea98393ff, deletion_comment = False)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/UnityCompositorInterface.cs
chrisfromwork: Do you have any examples for how to do this?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/CompositionManager.cs
chrisfromwork: this is a problem with the memory that is handed around, in the way it exists today we don't end up with a terminated string.

chrisfromwork: i think your string builder suggestion will fix this problem.

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
chrisfromwork: OutputDebugString already only writes things to the console in debug mode, but i guess it still may result in string allocation, ill add it to a _DEBUG stub

tools/Scripts/CopyPluginsToUnity.ps1
chrisfromwork: This is beyond the scope of this change, copying opencv requires meta files for opencv_*d.dll files which we don't currently have. If there ever becomes a need to debug in this manner, i will change things, but for now I am going to just enable copying debug compositor flags.

PR #218, opened 2019-09-16
Enable nightly builds

chrisfromwork:
This duplicates our existing ci-native.yam template to create a nightly build.

### DIFF
--- a/tools/ci/ci-nightly.yaml
+++ b/tools/ci/ci-nightly.yaml
@@ -1,10 +1,15 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See LICENSE in the project root for license information.
 
-# MixedReality-SpectatorView build native content
+# MixedReality-SpectatorView nightly build
 
-# Scheduled run, always run to monitor churn in external dependencies
-schedules:
+# Give a unique name to the build each time it runs
+name: ci-nightly-$(SourceBranchName)-$(Date:yyyyMMdd)-$(Rev:r)
+
+variables:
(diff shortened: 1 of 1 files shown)


tools/ci/ci-nightly.yaml
```diff
@@ -0,0 +1,43 @@
...
+# Scheduled run, always run to monitor churn in external dependencies
+schedules:
+- cron: "0 0 * * *"
+  displayName: Daily Midnight Build
+  branches:
+    include:
+    - master
+  always: true
+
+# Trigger CI on push changes
```
wiwei: Nit: It feels like this comment and the trigger of none don't seem to match up - i.e. it seems like this code/comment do match up:

https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#opting-out-of-ci-builds

tools/ci/ci-nightly.yaml
```diff
@@ -0,0 +1,43 @@
...
+trigger: none
+
+# Do not trigger CI on PRs
+pr: none
+
+# Give a unique name to the build each time it runs
+name: ci-nightly-$(SourceBranchName)-$(Date:yyyyMMdd)-$(Rev:r)
+
+variables:
```
wiwei: Nit: In general stick variables at the top so that you can see the key bits of data going into each YAML file (i.e. almost like function signature)

https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#set-variables-in-pipeline

In this case, I guess it's LESS interesting because UnityVersion is more or less a const

PR #226, opened 2019-09-26
Fix adding submodules to build repo.

chrisfromwork:
We're seeing errors related to repos already existing on agents. This should hopefully fix any of these issues.


chrisfromwork:
Merging to validate with CI

Issue #230, opened 2019-09-27
Extreme FPS drop when Spectator connects on HoloLens 2


ALaurain:
Hello, I am developing an application for the HoloLens2 using the new Spectator View submodule. I have tried building to iOS and Android, and have tested both the QR Code and Azure Spatial Anchors features. In all these scenarios I am able to view my holograms on the spectating device. However, no matter what changes I make, the FPS always drops to around 10 on the HoloLens when a spectator connects. When I kill the spectating app, the FPS jumps back to normal (around 40-45). Are there any ideas around what could be happening in my application? Is this a known issue that has happened to others? Any help would be greatly appreciated

To elaborate: I have the "automatically Broadcast All Gameobjects" checkbox set to false, and I have the GameObjectHierarchyBroadcaster script assigned to my "WorldRoot" gameobject. This object is the parent to all the holograms I want networked across to the spectator
I also noticed that many of my shaders (Microsoft standard shaders) are not rendering correctly on the spectating device. I did a full scene build onto ios and Android and found that the shaders are supported when the scene is built to the device directly, but when they are networked across, they break entirely

arsi0001:
Had also an project, where I should use the inspector. Same story → if broadcasting all game objects I got 1 fps, if broadcasting only certain objects I got 10 fps. Would also be interested in a solution/fix

chrisfromwork:
We are in the process of assessing tooling for better tuning these synchronization experiences. I am going to tag this issue as next release to prioritize improving things.

Currently, there are a few items floating around that may or may not prove useful:
1) In addition to placing GameObjectHierarchyBroadcasters in your scene, you can also tag items to not be synchronized with "SpectatorViewHidden" and "SpectatorViewChildrenHidden" tags. https://github.com/microsoft/MixedReality-SpectatorView/blob/f9ff8f43f45d88b25b94854288f59797aae93ea8/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/TransformBroadcaster.cs#L15

2) We have some old performance monitoring logic that may be tapped into via the video camera filming experience to see in the Unity editor the performance of the HoloLens experience. We will likely expand on this to try and improve the overall understanding of what game objects/components are causing the largest bottlenecks:
[...]

Issue #176, opened 2019-08-05
Investigate - Standard shader properties can fail to update material properties on spectator devices.


chrisfromwork:
This comes from community feedback. It still may require some investigation:

"I am trying out specator view for Hololens and Android , one of my objects has a material with standard shader with rendering mode fade. It fades on Hololens but not on Android.
Does anyone have any ideas ?
In my adroid build settings I got Auto graphics API checked off and OpenGLES3 in the APIS list."

At first glance, it may be related to the following issue:
https://forum.unity.com/threads/standard-material-shader-ignoring-setfloat-property-_mode.344557/

chrisfromwork:
We could add documentation around how to specify custom shader properties for synchronizing

We could add a material update extension point for different shader keywords

It seems like the standard shader should be supported by default

chrisfromwork:
It appears that property reporting works correctly. If anyone hits this issue, they may be able to unblock themselves by including their desired assets in a scene that is included in the application.

chrisfromwork:
We've added troubleshooting documentation for this issue. If you include the HoloLens scene in your android/ios builds, you should be able to pick up the shader variant you need based on materials referenced in this scene. Otherwise, you likely need to mess around with different graphics settings.

PR #241, opened 2019-10-10
Update performance reporting and improve synchronization documentation

chrisfromwork:
This review addresses the following Issues:
https://github.com/microsoft/MixedReality-SpectatorView/issues/176
https://github.com/microsoft/MixedReality-SpectatorView/issues/230

In this review:
1) We have documentation calling out how to fix build issues around transparent materials using the standard shader.
2) StateSynchronizationPerformanceMonitor has been updated to support custom event measures. If developers create custom component broadcasters, they can now leverage this measure functionality without needing to edit an internal enum.
3) StateSynchronizationPerformanceWindow can now read and display these new event types.
4) By default, performance reporting functionality does not execute unless it has been turned on by the StateSynchronizationPerformanceWindow. This was done to improve performance/avoid measuring event durations when no performance reporting was being conducted.
5) ComponentBroadcaster now contains a component name property that allows reporting performance events for some common events (frame update times, etc). Note, this property is opt-in/not required and should not result in compile breaks.
6) StateSynchronizationPerformanceParameter properties are now cached and only update on hierarchy changes. Previously, obtaining these inheritance properties always searched the scene hierarchy for StateSynchronizationPerformanceParameter components. Caching these values improves performance.
7) Material property values are now cached each frame update to avoid repeated calls to obtain properties from the material.
8) The MaterialBroadcaster now only caches properties that will be synchronized. TBD on whether this is acceptable. Looking for guidance from @matthejo 
9) A new SpectatorViewPerformance scene has been created that allows performance monitoring without needing to build the spectator view compositor native plugins.


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/AudioSource/AudioSourceBroadcaster.cs
```diff
@@ -21,6 +21,8 @@ public enum ChangeType
             Properties = 0x8
         }
 
+        public override string PerformanceComponentName => "AudioSourceBroadcaster";
```
matthejo: >"AudioSourceBroadcaster" [](start = 59, length = 24)

Is there a reason we can't use .GetType().Name for these? If not, it would be nice to use nameof(AudioSourceBroadcaster) so that renaming the component will automatically update these references too. Same goes for all of these overrides.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
```diff
@@ -101,7 +101,7 @@ public override void Read(SocketEndpoint sendingEndpoint, BinaryReader message)
                 TextMeshObserver.fontSizeMax = message.ReadSingle();
                 TextMeshObserver.fontSizeMin = message.ReadSingle();
                 TextMeshObserver.fontStyle = (FontStyles)message.ReadInt32();
-                TextMeshObserver.fontWeight = message.ReadInt32();
+                TextMeshObserver.fontWeight = (FontWeight)message.ReadInt32();
```
andreiborodin: Are we updating our TMP dependency here?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceMonitor.cs
```diff
@@ -2,89 +2,156 @@
...
             }
 
-            for (int i = 0; i < PeriodsToAverageOver; i++)
+            if (!dictionary.TryGetValue(eventName, out var stopwatch))
             {
-                previousSpentTimes[i] = new float[FeatureCount];
-                previousActualTimes[i] = new float[FeatureCount];
+                stopwatch = new Stopwatch();
+                dictionary.Add(eventName, stopwatch);
             }
```
andreiborodin: If we reuse stopwatch, this means we can't do event measurement re-entrancy?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceMonitor.cs
```diff
@@ -2,89 +2,156 @@
...
+                }
+            }
+        }
+
+        public void SetDiagnosticMode(bool enabled)
+        {
+            StateSynchronizationPerformanceParameters.EnablePerformanceReporting = enabled;
         }
 
         private struct TimeScope : IDisposable
```
andreiborodin: If we are on 7.3, we can use ref struct, I believe

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/ComponentBroadcaster.cs
```diff
@@ -133,93 +131,55 @@ public void OnFrameCompleted(SocketEndpointConnectionDelta connectionDelta)
...
 
-                        endpointsNeedingCompleteChanges.Clear();
-                        endpointsNeedingDeltaChanges.Clear();
-
-                        filteredEndpointsNeedingCompleteChanges.Clear();
-                        filteredEndpointsNeedingDeltaChanges.Clear();
-                        filteredEndpointsNotNeedingDeltaChanges.Clear();
-
-                        foreach (SocketEndpoint endpoint in connectionDelta.AddedConnections)
+                        if (TransformBroadcaster != null)
```
matthejo: >if (TransformBroadcaster != null) [](start = 24, length = 33)

Have you seen a case where the TransformBroadcaster becomes null here? It's already checked for null up a few lines, so it would have to be OnFrameCompleted or BeginUpdatingFrame that destroys it

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceMonitor.cs
```diff
@@ -2,89 +2,156 @@
...
-    internal class StateSynchronizationPerformanceMonitor : Singleton<StateSynchronizationPerformanceMonitor>
-    {
-        private const int PeriodsToAverageOver = 5;
-        private const int FeatureCount = (int)StateSynchronizationPerformanceFeature.Count;
-        private Stopwatch[] stopwatches;
-        private float[][] previousSpentTimes;
-        private float[][] previousActualTimes;
-        private int currentPeriod = 0;
+        private Dictionary<string, Dictionary<string, Stopwatch>> eventStopWatches = new Dictionary<string, Dictionary<string, Stopwatch>>();
+        private Dictionary<string, Dictionary<string, int>> eventCounts = new Dictionary<string, Dictionary<string, int>>();
```
matthejo: For both readability and performance, would prefer if there was a struct TimerKey with two fields (ComponentName and EventName) that can be used as the key for both of these dictionaries. This would avoid the extra tier of allocations for all of the internal dictionaries, and would make it easier to understand lookups when reading the code.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/AudioSource/AudioSourceBroadcaster.cs
chrisfromwork: nameof required a singleton definition. I will try GetType().Name

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/ComponentBroadcaster.cs
chrisfromwork: I'll remove this, this was primarily a check to make sure a TransformBroadcaster referenced itself. I've never seen this fired.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceMonitor.cs
chrisfromwork: if we think supporting re-entrancy is important we can change the value definition of these dictionarys to a stop watch + flag. That going said, I'm not sure how we would choose to support re-stopping a stop watch in an intuitive manner once reentrancy occurs.

chrisfromwork: We could build some sort of reference counting compared to running flag.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceParameters.cs
```diff
@@ -68,75 +83,138 @@ internal class MaterialPropertyPollingFrequency
...
+            // Stop the timer before calling parent function
+            return parentParameters.ShouldUpdateMaterialProperty(materialProperty);
+        }
+
+        public bool IsShaderSupported(string shader)
+        {
+            if (ShaderKeywords == PollingFrequency.UpdateContinuously)
+            {
+                return true;
             }
```
matthejo: I don't think this is a correct usage of ShaderKeywords. ShaderKeywords is meant to control how frequently the shaderKeywords of a material are synced (if those keywords could possibly change), not whether the entire shader is supported.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceParameters.cs
```diff
@@ -68,75 +83,138 @@ internal class MaterialPropertyPollingFrequency
...
+                    }
                 }
+            }
 
-                return materialProperties == PollingFrequency.UpdateContinuously;
+            // Stop the timer before calling parent function
+            return parentParameters.ShouldUpdateMaterialProperty(materialProperty);
+        }
+
+        public bool IsShaderSupported(string shader)
```
matthejo: >IsShaderSupported [](start = 20, length = 17)

Is the point of this method to filter out shaders that have no material properties on them? I guess I don't understand what "supported" means in this case. Maybe a more-descriptive name or doc comment would make this more clear.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/Demo/ChangeColor.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.SpectatorView
+{
+    public class ChangeColor : MonoBehaviour
```
matthejo: >ChangeColor [](start = 17, length = 11)

Is it necessary for this to be a public class in the main project? If so, I think it needs some documentation about what it's for.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
chrisfromwork: no, will revert

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/Demo/ChangeColor.cs
chrisfromwork: no, this can be internal

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceMonitor.cs
chrisfromwork: I think for now well leave this as is, in the future we can invest more in reentrancy support as needed.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceMonitor.cs
chrisfromwork: i'm going to go ahead and define this as a struct for now in case more variables exist in the future

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationObserver.cs
```diff
@@ -30,10 +32,10 @@ public class StateSynchronizationObserver : NetworkManager<StateSynchronizationO
         [SerializeField]
         protected int port = 7410;
 
-        private double[] averageTimePerFeature;
         private const float heartbeatTimeInterval = 0.1f;
         private float timeSinceLastHeartbeat = 0.0f;
         private HologramSynchronizer hologramSynchronizer = new HologramSynchronizer();
+       private StateSynchronizationPerformanceMonitor.ParsedMessage lastPerfMessage = new StateSynchronizationPerformanceMonitor.ParsedMessage(false, null, null);
```
matthejo: >        [](start = 0, length = 7)

Nit: missing space here

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceParameters.cs
chrisfromwork: I want to prevent iterating over all the known properties for each game object when the performance parameters don't support any continuous property synchronization. WIth that in mind, I have changed how this is defined for legibility/what I believe to be a more accurate flag for material synchronization supported.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/GlobalShaderPropertiesBroadcaster.cs
```diff
@@ -33,35 +33,38 @@ private void InitializeValues()
 
         public void OnFrameCompleted(SocketEndpointConnectionDelta connectionDelta)
         {
-            if (assetCache?.CustomGlobalShaderProperties != null)
+            using (StateSynchronizationPerformanceMonitor.Instance.MeasureEventDuration("GlobalShaderPropertiesBroadcaster", "OnFrameCompleted"))
```
matthejo: >"GlobalShaderPropertiesBroadcaster" [](start = 88, length = 35)

Could use nameof for both of these strings

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/MaterialsBroadcaster.cs
```diff
@@ -69,13 +71,20 @@ private MaterialPropertyAsset[] GetCachedMaterialProperties(int materialIndex, M
             return cachedMaterialPropertyAccessors[materialIndex];
         }
 
+        private bool ShouldAssessAnyPropertyAccessors(StateSynchronizationPerformanceParameters performanceParameters)
+        {
+            return !(!performanceParameters.HasCustomMateriaPropertyPollingFrequencies &&
+                performanceParameters.ShaderKeywords == StateSynchronizationPerformanceParameters.PollingFrequency.UpdateOnceOnStart &&
+                performanceParameters.MaterialProperties == StateSynchronizationPerformanceParameters.PollingFrequency.UpdateOnceOnStart);
```
matthejo: I think I'd find this easier to parse if the first ! was expanded over the rest of the conditions, e.g. "performanceParameters.HasCustomMateriaPropertyPollingFrequencies || performanceParameters.ShaderKeywords != StateSynchronizationPerformanceParameters.PollingFrequency.UpdateOnceOnStart || performanceParameters.MaterialProperties != StateSynchronizationPerformanceParameters.PollingFrequency.UpdateOnceOnStart

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationSceneManager.cs
```diff
@@ -340,93 +340,109 @@ private IEnumerator RunEndOfFrameUpdates()
             {
                 yield return new WaitForEndOfFrame();
 
-                if (framesToSkipForBuffering == 0)
+                using (StateSynchronizationPerformanceMonitor.Instance.MeasureEventDuration("StateSynchronizationSceneManager", "RunEndOfFrameUpdates"))
```
matthejo: >"StateSynchronizationSceneManager" [](start = 92, length = 34)

Could use nameof for all of these strings

Issue #242, opened 2019-10-11
local spatial alignment isn't cleaned up correctly on network disconnect


chrisfromwork:
Repro:
Run spectator view on an android device, connect to a hololens device or the editor running as the hololens, disrupt the network connection while the marker visual is displaying on device.

Expected:
the marker visual is hidden

Observed:
the marker visual remains/the spatial localization continues to run.

It looks like we complete task completion sources stating that localization has completed, but we don't actually cancel the localization session. To do this we may need to expose some sort of cancel function that can be accessed via the currentLocalizationSession

PR #243, opened 2019-10-11
Reshow the network connection dialogue on a network disconnect

chrisfromwork:
This pull request addresses https://github.com/microsoft/MixedReality-SpectatorView/issues/133

In this change:
1) A method is exposed on SocketClient that can be called to set a flag that prevents any additional connection attempts.
2) A flag is exposed on SpectatorView.cs that allows users to choose whether to automatically attempt reconnects when running as spectators.

Observed when working on this change: https://github.com/microsoft/MixedReality-SpectatorView/issues/242

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Socketer/SocketerClient.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Socketer/SocketerClient.cs
@@ -564,7 +564,7 @@ namespace Microsoft.MixedReality.SpectatorView
             private bool threadShouldRun;
             // If a connection fails, wait this many ms
             private const int timeoutMS = 500;
-            private bool attemptConnection = true;
+            private volatile bool attemptConnection = true;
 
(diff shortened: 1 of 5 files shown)


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Socketer/SocketerClient.cs
```diff
@@ -553,6 +564,7 @@ private class SocketClient
             private bool threadShouldRun;
             // If a connection fails, wait this many ms
             private const int timeoutMS = 500;
+            private bool attemptConnection = true;
```
fibann: nit: it might be better to make this and `shouldConnect` volatile if they are touched by multiple threads. It won't make a lot of difference here since they seem to be settable only once, but might avoid bugs down the line if the code changes.

stephenatwork: +1 for volatile. volatile also functions as a memory barrier in c# which is super useful.

PR #247, opened 2019-10-17
Enable cancelling localization sessions

chrisfromwork:
This review addresses: https://github.com/microsoft/MixedReality-SpectatorView/issues/242

In this review:
1) ILocalizationSession is extended to have a Cancel function defined. This is beneficial because it eases cleanup when a network connection is lost.
2) SpatialLocalizationSession was created as a base class. This base class contains cancellation token sources for tearing down localization on cancel.
3) Dispatcher and DispatcherUnity were created to ease scheduling tasks on the ui thread.
4) We now lock when accessing the current localization session in SpatialCoordinateSystemManager.
5) SpatialCoordinateSystemManager now has a flag that states whether localization is running.
6) SpatialAlignmentVisual now displays to the user when localization is running.

### Notes:
This review introduces breaking changes to ISpatialLocalizationSession. Contributors felt there was a need to be able to cancel sessions, which led to this change.

### Migration Instructions:
1) Update custom implementations of ISpatialLocalizationSession to support a Cancel function.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsAndroidCoordinateService.cs
+++ b/src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsAndroidCoordinateService.cs
@@ -17,7 +17,8 @@ namespace Microsoft.MixedReality.SpatialAlignment
     internal class SpatialAnchorsAndroidCoordinateService : SpatialAnchorsCoordinateService
     {
         private long lastFrameProcessedTimeStamp;
-        private static bool initialized = false;
+        private static object initializeLock = new object();
+        private static TaskCompletionSource<object> initializeCompletionSource = null;
 
         /// <summary>
         /// Instantiates a new <see cref="SpatialAnchorsAndroidCoordinateService"/>.
@@ -31,23 +32,19 @@ namespace Microsoft.MixedReality.SpatialAlignment
         /// <inheritdoc/>
         protected override Task OnInitializeAsync()
         {
-            if (initialized)
+            lock (initializeLock)
             {
-                Debug.Log("SpatialAnchorsAndroidCoordinateService: session already initialized");
-                return Task.CompletedTask;
-            }
+                if (initializeCompletionSource != null)
+                {
+                    Debug.Log("SpatialAnchorsAndroidCoordinateService: initializeCompletionSource already initialized");
+                    return initializeCompletionSource.Task;
+                }
 
-            TaskCompletionSource<object> taskCompletionSource = new TaskCompletionSource<object>();
+                initializeCompletionSource = new TaskCompletionSource<object>();
+            }
 
             UnityAndroidHelper.Instance.DispatchUiThread(unityActivity =>
             {
-                if (initialized)
-                {
-                    Debug.Log("SpatialAnchorsAndroidCoordinateService: session already initialized");
-                    taskCompletionSource.SetResult(null);
-                    return;
-                }
-
                 try
                 {
                     // We should only run the java initialization once
@@ -55,18 +52,17 @@ namespace Microsoft.MixedReality.SpatialAlignment
                     {
                         cloudServices.CallStatic("initialize", unityActivity);
                         Debug.Log("SpatialAnchorsAndroidCoordinateService: session successfully initialized");
-                        initialized = true;
-                        taskCompletionSource.SetResult(null);
+                        initializeCompletionSource.SetResult(null);
                     }
                 }
                 catch (Exception ex)
                 {
                     Debug.LogError($"Exception throw initializing SpatialAnchorAndroidCoordinateService: {ex.ToString()}");
-                    taskCompletionSource.SetException(ex);
+                    initializeCompletionSource.SetException(ex);
                 }
             });
 
-            return taskCompletionSource.Task;
+            return initializeCompletionSource.Task;
         }
 
         /// <inheritdoc/>
--- a/src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsAndroidCoordinateService.cs
+++ b/src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsAndroidCoordinateService.cs
@@ -17,6 +17,7 @@ namespace Microsoft.MixedReality.SpatialAlignment
     internal class SpatialAnchorsAndroidCoordinateService : SpatialAnchorsCoordinateService
     {
         private long lastFrameProcessedTimeStamp;
+        private static bool initialized = false;
 
         /// <summary>
         /// Instantiates a new <see cref="SpatialAnchorsAndroidCoordinateService"/>.
@@ -30,16 +31,31 @@ namespace Microsoft.MixedReality.SpatialAlignment
         /// <inheritdoc/>
         protected override Task OnInitializeAsync()
         {
+            if (initialized)
+            {
+                Debug.Log("SpatialAnchorsAndroidCoordinateService: session already initialized");
+                return Task.CompletedTask;
+            }
+
             TaskCompletionSource<object> taskCompletionSource = new TaskCompletionSource<object>();
 
             UnityAndroidHelper.Instance.DispatchUiThread(unityActivity =>
             {
+                if (initialized)
+                {
+                    Debug.Log("SpatialAnchorsAndroidCoordinateService: session already initialized");
+                    taskCompletionSource.SetResult(null);
+                    return;
+                }
+
                 try
                 {
                     // We should only run the java initialization once
                     using (AndroidJavaClass cloudServices = new AndroidJavaClass("com.microsoft.CloudServices"))
                     {
                         cloudServices.CallStatic("initialize", unityActivity);
+                        Debug.Log("SpatialAnchorsAndroidCoordinateService: session successfully initialized");
+                        initialized = true;
                         taskCompletionSource.SetResult(null);
                     }
                 }
--- a/src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsCoordinateService.cs
+++ b/src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsCoordinateService.cs
@@ -324,6 +324,11 @@ namespace Microsoft.MixedReality.SpatialAlignment
                     await session.CreateAnchorAsync(cloudSpatialAnchor);
                     return new SpatialAnchorsCoordinate(cloudSpatialAnchor, spawnedAnchorObject);
                 }
+                catch(TaskCanceledException)
+                {
+                    Debug.Log("Create coordinate was cancelled.");
+                    UnityEngine.Object.Destroy(spawnedAnchorObject);
+                }
                 catch
                 {
                     UnityEngine.Object.Destroy(spawnedAnchorObject);
@@ -334,6 +339,8 @@ namespace Microsoft.MixedReality.SpatialAlignment
             {
                 ReleaseSessionStartRequest();
             }
+
+            return null;
         }
 
         private async Task<ISpatialCoordinate> TryCreateCoordinateEditorAsync(Vector3 worldPosition, Quaternion worldRotation, CancellationToken cancellationToken)
--- a/src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsCoordinateService.cs
+++ b/src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsCoordinateService.cs
@@ -324,11 +324,6 @@ namespace Microsoft.MixedReality.SpatialAlignment
                     await session.CreateAnchorAsync(cloudSpatialAnchor);
                     return new SpatialAnchorsCoordinate(cloudSpatialAnchor, spawnedAnchorObject);
                 }
-                catch(TaskCanceledException)
-                {
-                    Debug.Log("Create coordinate was cancelled.");
-                    UnityEngine.Object.Destroy(spawnedAnchorObject);
-                }
                 catch
                 {
                     UnityEngine.Object.Destroy(spawnedAnchorObject);
--- a/src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsCoordinateService.cs
+++ b/src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsCoordinateService.cs
@@ -334,8 +334,6 @@ namespace Microsoft.MixedReality.SpatialAlignment
             {
                 ReleaseSessionStartRequest();
             }
-
-            return null;
         }
 
         private async Task<ISpatialCoordinate> TryCreateCoordinateEditorAsync(Vector3 worldPosition, Quaternion worldRotation, CancellationToken cancellationToken)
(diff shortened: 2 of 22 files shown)


src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsAndroidCoordinateService.cs
```diff
@@ -30,16 +31,31 @@ public SpatialAnchorsAndroidCoordinateService(SpatialAnchorsConfiguration spatia
         /// <inheritdoc/>
         protected override Task OnInitializeAsync()
         {
+            if (initialized)
```
andreiborodin: What I would recommend is to cache the taskCompletionSource in a static variable, and perform a null check on it to see if it needs instantiation or not. If it doesn't, just return it's task. Also, you can wrap all of it in a lock.

src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsCoordinateService.cs
```diff
@@ -324,6 +324,11 @@ private async Task<ISpatialCoordinate> TryCreateCoordinateImplAsync(Vector3 worl
                     await session.CreateAnchorAsync(cloudSpatialAnchor);
                     return new SpatialAnchorsCoordinate(cloudSpatialAnchor, spawnedAnchorObject);
                 }
+                catch(TaskCanceledException)
```
andreiborodin: I would recommend making this a OperationCancelledException, which is the base of TaskCancelledException

src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpectatorView/SpatialAnchorsLocalizer.cs
```diff
@@ -90,17 +90,17 @@ private void Update()
             Updated?.Invoke();
         }
 
-        private class SpatialCoordinateLocalizationSession : DisposableBase, ISpatialLocalizationSession
+        private class SpatialCoordinateLocalizationSession : SpatialLocalizationSession
         {
-            public IPeerConnection Peer => peerConnection;
+            public override IPeerConnection Peer => peerConnection;
```
andreiborodin: Did you mean to add "///<inheritdoc/>" here as well?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/MarkerVisual/MarkerVisualDetectorSpatialLocalizer.cs
```diff
@@ -70,49 +71,59 @@ public LocalizationSession(MarkerVisualDetectorSpatialLocalizer localizer, Marke
...
             }
 
             /// <inheritdoc />
-            public async Task<ISpatialCoordinate> LocalizeAsync(CancellationToken cancellationToken)
+            public override async Task<ISpatialCoordinate> LocalizeAsync(CancellationToken cancellationToken)
             {
-                DebugLog($"Waiting for marker visual, CanBeCanceled:{cancellationToken.CanBeCanceled}, IsCancellationRequested:{cancellationToken.IsCancellationRequested}");
-                await Task.WhenAny(coordinateAssigned.Task, Task.Delay(-1, cancellationToken));
-                if (string.IsNullOrEmpty(coordinateId))
+                if (!defaultCTS.Token.CanBeCanceled)
```
andreiborodin: >defaultCTS.Token [](start = 21, length = 16)

Instead of exposing the CTS, I would recommend an auto property of type CancellationToken that you set with the CTS token as soon as you create it. This way, if the CTS is cancelled and disposed, the token can be still used in it's cancelled state instead of throwing objectdisposedexception

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemManager.cs
```diff
@@ -178,7 +205,7 @@ public Task<bool> RunRemoteLocalizationAsync(SocketEndpoint socketEndpoint, Guid
             if (remoteLocalizationSessions.TryGetValue(socketEndpoint, out var currentCompletionSource))
             {
                 DebugLog("Canceling existing remote localization session: {socketEndpoint.Address}");
-                currentCompletionSource.TrySetCanceled();
+                currentCompletionSource.TrySetResult(false);
```
andreiborodin: >TrySetResult [](start = 40, length = 12)

Why not cancel here?

chrisfromwork: No preference. I guess I thought this would be called from a user action, but I will just revert.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/WorldAnchor/WorldAnchorSpatialLocalizer.cs
```diff
@@ -63,47 +63,62 @@ public override bool TryCreateLocalizationSession(IPeerConnection peerConnection
...
-                }
+
+                return await Task.FromResult(spatialCoordinate);
             }
 
-            public void OnDataReceived(BinaryReader reader)
+            /// <inheritdoc />
+            public override void OnDataReceived(BinaryReader reader)
             {
             }
```
matthejo: If this is now virtual and we don't need to explicitly not have the base method happen, we can just get rid of this empty method now

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/PhysicalMarker/MarkerDetectorSpatialLocalizer.cs
```diff
@@ -46,26 +47,34 @@ public LocalizationSession(MarkerDetectorSpatialLocalizer localizer, MarkerDetec
...
                 }
 
                 return spatialCoordinate;
             }
 
-            public void OnDataReceived(BinaryReader reader)
+            /// <inheritdoc />
+            public override void OnDataReceived(BinaryReader reader)
             {
             }
```
matthejo: Can get rid of this override

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialLocalizationSession.cs
```diff
@@ -0,0 +1,47 @@
...
+            if (!this.IsDisposed &&
+                defaultCTS != null &&
+                defaultCTS.Token.CanBeCanceled)
+            {
+                defaultCTS.Cancel();
+            }
+        }
+
+        /// <inheritdoc />
+        public abstract void OnDataReceived(BinaryReader reader);
```
matthejo: >public abstract void OnDataReceived(BinaryReader reader); [](start = 8, length = 57)

I would just make this virtual with an empty body, since it's sometimes not needed by subclasses

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemParticipant.cs
```diff
@@ -27,7 +27,7 @@ public class SpatialCoordinateSystemParticipant : DisposableBase, IPeerConnectio
         private bool showDebugVisuals;
         private readonly TaskCompletionSource<ISet<Guid>> peerSupportedLocalizersTaskSource = new TaskCompletionSource<ISet<Guid>>();
 
-        public SocketEndpoint SocketEndpoint { get; }
+        public SocketEndpoint SocketEndpoint { get; set; }
```
matthejo: >public SocketEndpoint SocketEndpoint { get; set; } [](start = 8, length = 50)

Under what circumstances can the endpoint associated with a participant change?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemManager.cs
```diff
@@ -265,16 +276,17 @@ private void SendLocalizationCompleteCommand(SocketEndpoint socketEndpoint, bool
 
         private void OnConnected(SocketEndpoint endpoint)
         {
-            if (participants.ContainsKey(endpoint))
+            if (participants.TryGetValue(endpoint.Address, out var existingParticipant))
```
matthejo: >if (participants.TryGetValue(endpoint.Address, out var existingParticipant)) [](start = 12, length = 76)

Why the change to use the address? I'm confused about when this could happen and why it's not an error.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemParticipant.cs
chrisfromwork: we saw locally that Unity buttons ended up having two event calls on a single press based on some input configurations. This caused two socket endpoints to get created for the same ip address. TCPConnectionManager has been updated to prevent creating multiple end points with subsequent event calls, but other networking abstractions may not be as friendly.

Do you believe it is safe to key off ip addresses or can you imagine scenarios where the same ip address creates multiple endpoints for the same shared experience?

matthejo: I could certainly imagine scenarios where this would break, e.g. if we wanted two devices both connected over IPOverUSB to work (on different ports). They would both have 127.0.0.1 as their address but be bound to different ports and have different endpoints.

---
In reply to: [337232321](https://github.com/microsoft/MixedReality-SpectatorView/pull/247#discussion_r337232321) [](ancestors = 337232321)

src/SpectatorView.Unity/Assets/SpatialAlignment/Scripts/Utilities/Extensions.cs
```diff
@@ -182,9 +182,9 @@ public static Task AsTask(this CancellationToken cancellationToken)
         /// <param name="task">The task to await.</param>
         /// <param name="cancellationToken">The cancellation token to stop awaiting.</param>
         /// <returns>The task that can be awaited unless the cancellation token is triggered.</returns>
-        public static Task Unless(this Task task, CancellationToken cancellationToken)
+        public static async Task Unless(this Task task, CancellationToken cancellationToken)
         {
-            return Task.WhenAny(task, cancellationToken.AsTask());
+            await (await Task.WhenAny(task, cancellationToken.AsTask()));
```
andreiborodin: > await (await Task.WhenAny(task, cancellationToken.AsTask())); [](start = 11, length = 62)

Actually this could be just one await and teh function not async, but the bug was that there was no await at all in old version.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemManager.cs
```diff
@@ -48,11 +48,27 @@ public class SpatialCoordinateSystemManager : Singleton<SpatialCoordinateSystemM
         private const string LocalizationCompleteCommand = "LOCALIZEDONE";
         private readonly Dictionary<Guid, ISpatialLocalizer> localizers = new Dictionary<Guid, ISpatialLocalizer>();
         private readonly Dictionary<SocketEndpoint, TaskCompletionSource<bool>> remoteLocalizationSessions = new Dictionary<SocketEndpoint, TaskCompletionSource<bool>>();
-        private Dictionary<SocketEndpoint, SpatialCoordinateSystemParticipant> participants = new Dictionary<SocketEndpoint, SpatialCoordinateSystemParticipant>();
+        private Dictionary<string, SpatialCoordinateSystemParticipant> participants = new Dictionary<string, SpatialCoordinateSystemParticipant>();
         private HashSet<INetworkManager> networkManagers = new HashSet<INetworkManager>();
-        private ISpatialLocalizationSession currentLocalizationSession = null;
         private ITrackingObserver trackingObserver = null;
 
+        private object localizationLock = new object();
```
matthejo: >private [](start = 8, length = 7)

this field could also be readonly

src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsCoordinateService.cs
```diff
@@ -334,6 +334,8 @@ private async Task<ISpatialCoordinate> TryCreateCoordinateImplAsync(Vector3 worl
             {
                 ReleaseSessionStartRequest();
             }
+
+            return null;
```
matthejo: >return null; [](start = 12, length = 12)

This seems like unreachable code, how would this ever be hit?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemParticipant.cs
chrisfromwork: i will revert this now.

src/SpectatorView.Unity/Assets/SpatialAlignment/Scripts/Utilities/Extensions.cs
chrisfromwork: we tested this together locally and will leave this how it is.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemManager.cs
chrisfromwork: reverting

src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsCoordinateService.cs
chrisfromwork: product of removing a catch stub, fixing now.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemManager.cs
```diff
@@ -50,9 +50,25 @@ public class SpatialCoordinateSystemManager : Singleton<SpatialCoordinateSystemM
         private readonly Dictionary<SocketEndpoint, TaskCompletionSource<bool>> remoteLocalizationSessions = new Dictionary<SocketEndpoint, TaskCompletionSource<bool>>();
         private Dictionary<SocketEndpoint, SpatialCoordinateSystemParticipant> participants = new Dictionary<SocketEndpoint, SpatialCoordinateSystemParticipant>();
         private HashSet<INetworkManager> networkManagers = new HashSet<INetworkManager>();
-        private ISpatialLocalizationSession currentLocalizationSession = null;
         private ITrackingObserver trackingObserver = null;
 
+        private readonly object localizationLock = new object();
```
stephenatwork: Is there a reason to prefer a separate lock object? If this is only protecting currentLocalizationSession, would we lock that object directly instead?

chrisfromwork: locking on a null object throws exceptions. Since we want to populate currentLocalizationSession within the lock, it seems safer to lock on a separate, non-null object.

PR #249, opened 2019-10-18
Working around Socket bug that blocked communication after large asset bundles are sent.

mrbobbybobberson:


mrbobbybobberson:
@matthejo, how does this look?

PR #250, opened 2019-10-21
Adding Generic Observer Project

mrbobbybobberson:
Created a generic observer project that contains no built-in asset caches, but can receive asset
bundles.  It can be built for HoloLens, ARCore, or ARKit.  It has TextMeshPro sync'ing turned on.

Also changed mesh broadcaster/observer so a Unity error is not logged when sync'ing across a
null or missing mesh.


mrbobbybobberson:
@matthejo, what do you think of the project name?  It's currently "GenericSpectator", but here are some other options I thought of:

  * SpectatorView.GenericSpectator
  * SpectatorView.GenericObserver
  * SpectatorView.Spectator
  * SpectatorView.Observer

src/GenericSpectator/Assets/Scripts/PlatformDeviceBootstrapper.cs
```diff
@@ -0,0 +1,49 @@
+﻿using UnityEngine;
+
+public class PlatformDeviceBootstrapper : MonoBehaviour
```
mrbobbybobberson: >PlatformDeviceBootstrapper [](start = 13, length = 26)

@matthejo, I didn't want to have to build a different scene for each platform, so I made a single scene with a prefab instantiator for the specific platform's device (camera+goo) stuff.  Any concerns with this approach?

When the ARCore and/or ARKit SDK folders are hidden, the corresponding prefab has missing scripts in it, but as long as I don't instantiate the prefab, the console/error log doesn't get spammed with "missing script" warnings, so that's fortunate.

src/GenericSpectator/ProjectSettings/ProjectSettings.asset
```diff
@@ -0,0 +1,881 @@
...
+  productGUID: 674eb9d4c0e4c4e4691c31ed7ecfcebd
+  AndroidProfiler: 0
+  AndroidFilterTouchesWhenObscured: 0
+  AndroidEnableSustainedPerformanceMode: 0
+  defaultScreenOrientation: 0
+  targetDevice: 2
+  useOnDemandResources: 0
+  accelerometerFrequency: 60
+  companyName: Microsoft
+  productName: SpectatorView.GenericSpectator
```
mrbobbybobberson: >SpectatorView.GenericSpectator [](start = 15, length = 30)

@matthejo, OK with this product name?

src/GenericSpectator/ProjectSettings/ProjectSettings.asset
```diff
@@ -0,0 +1,881 @@
...
+  useHDRDisplay: 0
+  m_ColorGamuts: 00000000
+  targetPixelDensity: 30
+  resolutionScalingMode: 0
+  androidSupportedAspectRatio: 1
+  androidMaxAspectRatio: 2.1
+  applicationIdentifier:
+    Android: com.Microsoft.MixedReality.SpectatorView.GenericSpectator
+    iOS: com.Microsoft.MixedReality.SpectatorView.GenericSpectator
+  buildNumber: {}
```
mrbobbybobberson: @matthejo, OK with these app identifiers?

src/GenericSpectator/ProjectSettings/ProjectSettings.asset
```diff
@@ -0,0 +1,881 @@
...
+  incrementalIl2cppBuild: {}
+  allowUnsafeCode: 0
+  additionalIl2CppArgs: 
+  scriptingRuntimeVersion: 1
+  apiCompatibilityLevelPerPlatform:
+    Android: 6
+  m_RenderingPath: 1
+  m_MobileRenderingPath: 1
+  metroPackageName: SpectatorView.GenericSpectator
+  metroPackageVersion: 1.0.0.0
```
mrbobbybobberson: @matthejo, OK with this app name for HoloLens?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/MeshFilter/MeshFilterBroadcaster.cs
```diff
@@ -38,9 +38,9 @@ protected override void OnInitialized()
 
             meshFilter = GetComponent<MeshFilter>();
             assetId = AssetService.Instance.GetMeshId(meshFilter.sharedMesh);
-            if (assetId == Guid.Empty)
+            if ((meshFilter.sharedMesh != null) && (assetId == Guid.Empty))
```
mrbobbybobberson: >if [](start = 12, length = 2)

In the Build2019 demo app, there's one mesh filter whose mesh is missing.  That would make the MeshFilterBroadcaster log an error about the asset not being the asset cache.  This change makes it sync across null/missing meshes without treating them as errors.

Make sense?  Want me to keep this change in with my PR or revert it?

src/GenericSpectator/Assets/Scenes/SpectatorView.GenericObserver.unity
```diff
@@ -0,0 +1,252 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
```
mrbobbybobberson: Hmm... I should probably change the name of this scene to match the project name better.  Perhaps "SpectatorView.GenericSpectator" assuming the project name stays of that form.

src/GenericSpectator/Assets/Scripts/PlatformDeviceBootstrapper.cs
chrisfromwork: FWIW, this is how we have modeled some other platform specific components. ITrackingObservers are instantiated in this manner.

src/GenericSpectator/ProjectSettings/ProjectSettings.asset
chrisfromwork: We should add a .Unity extension to this project's folder naming. long term src may contain other middleware code bases

andreiborodin:
Do we need to check-in the imported TMP assets?

src/GenericSpectator/Assets/Scripts/PlatformDeviceBootstrapper.cs
andreiborodin: Missing namespace.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetService.cs
```diff
@@ -108,7 +108,7 @@ public bool AttachMeshFilter(GameObject gameObject, Guid assetId)
             ComponentExtensions.EnsureComponent<MeshRenderer>(gameObject);
 
             Mesh mesh = meshAssets.GetAsset(assetId);
-            if (mesh != null)
+            if ((mesh != null) || (assetId == Guid.Empty))
```
andreiborodin: >if ((mesh != null) || (assetId == Guid.Empty)) [](start = 12, length = 46)

So if the assetId is a Guid.Empty, we don't have to check the cache, and furthermore, if we aren't treating that as an error when sending across, how do we handle it here, do we still attach a meshfilter, do we skip attaching but return true, or do we return false?

src/GenericSpectator/Assets/Scripts/PlatformDeviceBootstrapper.cs
```diff
@@ -0,0 +1,49 @@
+﻿using UnityEngine;
```
chrisfromwork: nit: new files should have the microsoft open source copy right header:

src/GenericSpectator/Assets/Scripts/PlatformDeviceBootstrapper.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    /// The device prefab for the iOS platform.
+    /// </summary>
+    [Tooltip("The device prefab for the iOS platform.")]
+    [SerializeField]
+    protected GameObject iOSDevicePrefab;
+
+    protected void Awake()
+    {
+        GameObject devicePrfab;
```
chrisfromwork: nit: devicePrefab

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/MeshFilter/MeshFilterBroadcaster.cs
chrisfromwork: what's the benefit of creating/syncing null/missing meshes? Are we hoping that logging an unknown mesh on both the user and spectator devices makes fixing the problem easier to troubleshoot?

mrbobbybobberson:
That's a good question.  If this project is built without them, the generic spectator won't be able to render TMP visuals, and since TMP is so popular, I wouldn't want someone to accidentally forget to include it.

If y'all want, I could put TMP in .gitignore and omit it from the repo, but then I'd probably want to make a build step to emit a warning if someone tries to build without it and perhaps put a dummy TMP object in the scene hierarchy to get Unity to show its "Import TMP Essentials" dialog.

Thoughts?

---
In reply to: [544714011](https://github.com/microsoft/MixedReality-SpectatorView/pull/250#issuecomment-544714011) [](ancestors = 544714011)

src/GenericSpectator/Assets/Scripts/PlatformDeviceBootstrapper.cs
mrbobbybobberson: Haha, I had to read that a couple times before I noticed... maybe I should up the font size on my monitor ;-).

Thanks for the catch!

---
In reply to: [337248663](https://github.com/microsoft/MixedReality-SpectatorView/pull/250#discussion_r337248663) [](ancestors = 337248663)

src/GenericSpectator/Assets/Scripts/PlatformDeviceBootstrapper.cs
mrbobbybobberson: Cool.  Good to know!

---
In reply to: [337246014](https://github.com/microsoft/MixedReality-SpectatorView/pull/250#discussion_r337246014) [](ancestors = 337246014)

src/GenericSpectator/Assets/Scripts/PlatformDeviceBootstrapper.cs
mrbobbybobberson: Thanks!

---
In reply to: [337248381](https://github.com/microsoft/MixedReality-SpectatorView/pull/250#discussion_r337248381) [](ancestors = 337248381)

src/GenericSpectator/Assets/Scripts/PlatformDeviceBootstrapper.cs
mrbobbybobberson: Thanks!

---
In reply to: [337246998](https://github.com/microsoft/MixedReality-SpectatorView/pull/250#discussion_r337246998) [](ancestors = 337246998)

src/GenericSpectator/ProjectSettings/ProjectSettings.asset
mrbobbybobberson: Thanks!

---
In reply to: [337246501](https://github.com/microsoft/MixedReality-SpectatorView/pull/250#discussion_r337246501) [](ancestors = 337246501)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/MeshFilter/MeshFilterBroadcaster.cs
mrbobbybobberson: I made it so it *wouldn't* log if someone tried to sync a null/missing mesh.

Since a null mesh is a valid state, and since a missing resource is typically already warned about by Unity, it doesn't seem like SpectatorView's roll to log in either of those cases.  SpectatorView's roll should just be to make the observer see the same visuals as the broadcaster, and sync'ing over the null mesh accomplishes that.

... but in truth, I only noticed this because I was getting a confusing error message when running the Build2019Demo scene and wanted to make sure I hadn't broken something.  If y'all would prefer to keep the error, but reword it to also mention null or missing meshes, I could do that instead, and I could remove the offending MeshFilter/GameObject from the Build2019Demo scene to make the error go away instead.

Thoughts?

---
In reply to: [337250788](https://github.com/microsoft/MixedReality-SpectatorView/pull/250#discussion_r337250788) [](ancestors = 337250788)

src/GenericSpectator/Assets/Scenes/SpectatorView.GenericObserver.unity
chrisfromwork: if its a standalone project, i've seen "Main" used as well

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetService.cs
mrbobbybobberson: The intention here was to treat "Guid.Empty" to mean "the null asset".

I was on the fence though about whether I should still let the call to GetAsset happen even when assetId is Guid.Empty or just short-circuit it to null here.

In the end, I decided to let the GetAsset call happen, because I thought perhaps in the future, Guid.Empty and the null asset might be treated in a special way, but I'm open to short-circuiting it here instead.

Since sync'ing across a null mesh is now a supported scenario, I figured it should return true.

---
In reply to: [337248018](https://github.com/microsoft/MixedReality-SpectatorView/pull/250#discussion_r337248018) [](ancestors = 337248018)

src/GenericSpectator/Assets/Scenes/SpectatorView.GenericObserver.unity
mrbobbybobberson: That's a good idea.  I'll probably switch to that.

---
In reply to: [337259015](https://github.com/microsoft/MixedReality-SpectatorView/pull/250#discussion_r337259015) [](ancestors = 337259015)

matthejo:
If the intent is for it to be a mobile spectator (e.g. not a HoloLens generic spectator), maybe "SpectatorView.GenericMobileSpectator" for the name? One other naming idea I had instead of "Generic" would be "Universal", e.g. SpectatorView.UniversalMobileSpectator. Thoughts?

---
In reply to: [544701897](https://github.com/microsoft/MixedReality-SpectatorView/pull/250#issuecomment-544701897) [](ancestors = 544701897)

mrbobbybobberson:
I could see this (and would kinda like to see this) also built for HoloLens, though I don't know if I'll have time to validate that scenario (particularly around coordinate space sync'ing), so I'd prefer to keep "Mobile" out of the name.

Ooh, I think I like "Universal" slightly better than "Generic", so I'll consider "SpectatorView.UniversalSpectator.Unity" the top contender for now.  Any other votes?

---
In reply to: [544735442](https://github.com/microsoft/MixedReality-SpectatorView/pull/250#issuecomment-544735442) [](ancestors = 544735442,544701897)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/MeshFilter/MeshFilterBroadcaster.cs
mrbobbybobberson: As noted elsewhere, I'm reverting this change, and we can consider it as part of a separate PR.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetService.cs
mrbobbybobberson: As discussed elsewhere, I reverted this change, and we can consider it in another PR.

src/SpectatorView.UniversalSpectator.Unity/Assets/ARKit-Unity-Plugin
```diff
@@ -0,0 +1 @@
+../../.././external/ARKit-Unity-Plugin
```
matthejo: I think you should name this UnityARKitPlugin, as that's the directory name that the ASA script tries to hide / show when changing to the iOS platform

mrbobbybobberson: Oh, you're right.  Thanks for the catch!  I was a bit confused, because the ASA plugin used one name, but the "add dependencies" script used another.  Perhaps a separate PR should be to make the "add dependencies" script and SpectatorView.Example.Unity use the same name as ASA.

Issue #216, opened 2019-09-11
Documentation should provide tips for fixing shader compile errors for android and iOS


chrisfromwork:
This seems to be the one of the major pain points when beginning extending experiences to android and iOS. Providing tips that are known amongst the contributor team could help developers struggling to prepare their project for android/iOS.

PR #261, opened 2019-10-28
Add troubleshooting item to ease the process of fixing shaders for iOS/Android

chrisfromwork:
This review addresses: https://github.com/microsoft/MixedReality-SpectatorView/issues/216


doc/SpectatorView.Setup.md
```diff
@@ -224,4 +224,15 @@ Scraping the entire Unity scene for content updates on the HoloLens device can b
...
+![Marker](images/FixTransparency.png)
+
+### __Issue:__ Shaders don't compile for Android or iOS Unity Players
+Shaders originally created to run on HoloLens may not immediately compile for Android or iOS. One common issue is that shader model 5.0 is not supported by OpenGL. To hide DirectX11 shader model 5.0 logic, you can use the below `#if defined(SHADER_API_D3D11)` (Note: In some instances, shaders may not appear correctly in the editor but will compile correctly for iOS and Android). For more information on shader target levels, see [here](https://docs.unity3d.com/Manual/SL-ShaderCompileTargets.html).
+
+`#if defined(SHADER_API_D3D11)`
+
+`#pragma target 5.0`
+
+`#endif`
```
matthejo: Have you tested this out? At least in an older version of Unity, #pragma target values were processed by the preprocessor without parsing any #if blocks, and you couldn't conditionally change the target. Maybe they've fixed this, though?

chrisfromwork: i think this is fixed, at least at the time of compilation it seems to work. I'm not sure if it ever works in the editor.

Issue #47, opened 2019-06-27
Update state synchronization logic to have asset cache asset ids created in a deterministic manner.


chrisfromwork:
Right now asset caches need to be created on the same developer machine and shared. This is because the Guid id that is created for each asset isn't created in a deterministic manner. Changing this behavior will result in a better developer experience for multi-person teams. It will also ease the process of working between a PC and mac.

PR #264, opened 2019-10-31
Add deterministic asset id generation process for asset caches.

chrisfromwork:
This review addresses: https://github.com/microsoft/MixedReality-SpectatorView/issues/47. It contains the following changes:

1) A new AssetId class has been created to use as a key for AssetCacheEntries. This type contains a GUID and long is based on Asset file GUIDs and file identifiers
2) Asset caches are updated to use this new AssetId key

### Notes:
StringGuid keys for asset caches have been updated to AssetIds so that they can be deterministically created on different computers. This will break old AssetCaches.

### Migration Instructions:
1. Delete asset caches in your Generated.StateSynchronization.AssetCaches\Resources folder (these will be the .asset files).
1. Call `Spectator View -> Update All Asset Caches` in your project to regenerate asset caches.
2. Recompile all player flavors (UWP, iOS, Android) of your application.

### DIFF
--- a/samples/Build2019Demo.Unity/Assets/Generated.StateSynchronization.AssetCaches/Resources/SpriteAssetCache.asset
+++ b/samples/Build2019Demo.Unity/Assets/Generated.StateSynchronization.AssetCaches/Resources/SpriteAssetCache.asset
@@ -14,176 +14,297 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   assets:
   - AssetId:
-      m_storage: 74fae14d-39c0-497e-ae6d-72f1f2084231
-    Asset: {fileID: 21300000, guid: d651fc00b471cfe44834a222e9433461, type: 3}
+      guid:
+        m_storage: 00000000-0000-0000-f000-000000000000
+      fileIdentifier: 10905
+    Asset: {fileID: 10905, guid: 0000000000000000f000000000000000, type: 0}
   - AssetId:
-      m_storage: f42bb34d-53dc-481f-bcc7-9823e4c197e7
-    Asset: {fileID: 21300000, guid: 6deea620179caad4d97d51e0e6e03010, type: 3}
-  - AssetId:
-      m_storage: 30df0952-4da8-4b65-a4bf-9e2bd95db273
-    Asset: {fileID: 21300000, guid: 84643a20fa6b4fa7969ef84ad2e40992, type: 3}
-  - AssetId:
-      m_storage: e52d6769-916e-45d4-83f9-9e60a8d5fa94
-    Asset: {fileID: 21300000, guid: acc34040a66fe4170bc8885268860cfe, type: 3}
-  - AssetId:
-      m_storage: f842e9e4-deba-495b-ae69-a3b0ed76f8fd
-    Asset: {fileID: 21300000, guid: b2c8cdd0abb099e43b24fe3c1944e166, type: 3}
-  - AssetId:
-      m_storage: 091b7262-8f5c-43b5-b8a4-401a6a083eb3
-    Asset: {fileID: 21300000, guid: 3ba76ce0e2c9a7c45801facdc7e24384, type: 3}
-  - AssetId:
-      m_storage: 395056aa-7f64-491e-b73b-456d289b9de3
-    Asset: {fileID: 21300000, guid: 61438131b4fdac446a91f664f1770aad, type: 3}
-  - AssetId:
-      m_storage: 11b0cc54-384a-4363-a516-78cb37b3697c
-    Asset: {fileID: 21300000, guid: 8a0da39193021fa4cb9b71d883d30b97, type: 3}
-  - AssetId:
-      m_storage: c48ae6f2-9a63-4ebb-8beb-f609a0f00b8b
-    Asset: {fileID: 21300000, guid: 2e1868a15829536459f5d0022d6ec0b5, type: 3}
-  - AssetId:
-      m_storage: c7fb622c-deb9-4e38-b041-36ae1855cec4
-    Asset: {fileID: 21300000, guid: 4af5db224464bf34cb8fdc70ee470b65, type: 3}
-  - AssetId:
-      m_storage: e51c16d3-814c-4e17-bc17-b7170e831b05
-    Asset: {fileID: 21300000, guid: f2237692de9624a4692e4b0440c5f476, type: 3}
-  - AssetId:
-      m_storage: 207bf807-be6c-4f35-9c67-1c8a708ff975
-    Asset: {fileID: 21300000, guid: a7d0f692b982a754b8ae2353d5e88d64, type: 3}
-  - AssetId:
-      m_storage: b12a4dc8-90a6-4f1b-9a81-2497701f2eed
-    Asset: {fileID: 21300000, guid: 1520ffe22e0382949b34d64a73900077, type: 3}
-  - AssetId:
-      m_storage: 8a3a318c-6240-48a7-aff9-7a34ae5f6197
-    Asset: {fileID: 21300000, guid: c2c00ef21cc44bcfa09695879e0ebecd, type: 3}
-  - AssetId:
-      m_storage: 1b276a3c-1b19-4075-8a07-3dd84c128283
-    Asset: {fileID: 21300000, guid: 0b431d73cc837214fbf67cab3535f975, type: 3}
-  - AssetId:
-      m_storage: b7ddc170-8d2d-4321-a232-24c232274894
-    Asset: {fileID: 21300000, guid: 739d0ad3b6c10dc49b8a45caf9002249, type: 3}
-  - AssetId:
-      m_storage: 4b38f031-3567-4874-9c21-28e2d6d43b06
-    Asset: {fileID: 21300000, guid: d867b2e361e90944b8b1ee90338f4350, type: 3}
-  - AssetId:
-      m_storage: f936f98c-201f-467b-a2ee-9eb78430b155
-    Asset: {fileID: 21300000, guid: 0ac4e434e8bf54a418c7e6ff28cdf07b, type: 3}
-  - AssetId:
-      m_storage: a54d1215-1350-4093-aa43-b98fc6866ae3
-    Asset: {fileID: 21300000, guid: 6344c4749a5854a4fb033def626a8547, type: 3}
-  - AssetId:
-      m_storage: 2970e75b-5dae-4501-8b53-f3f0709fd69e
-    Asset: {fileID: 21300000, guid: 0ea40b842547d784082ebf2698d3f08b, type: 3}
-  - AssetId:
-      m_storage: 6b8b9634-a79a-4c49-b27a-c81f1e827dac
-    Asset: {fileID: 21300000, guid: 6588fe8449020f74488b8b667ce2f435, type: 3}
-  - AssetId:
-      m_storage: 88e14552-363c-4d53-8cd5-9d77133113c2
-    Asset: {fileID: 21300000, guid: 3cf132d4fa07f4f0cb2883499c5c7dd0, type: 3}
-  - AssetId:
-      m_storage: 9e196a5d-d80a-4431-aab5-8dbeb467cf6c
-    Asset: {fileID: 21300000, guid: 96d001f4572ca3e4485a62e1a049524e, type: 3}
-  - AssetId:
-      m_storage: dc63b986-89db-433d-8699-c55726572b4f
-    Asset: {fileID: 21300000, guid: e331c3759dc6afb4086a58238fc6934c, type: 3}
-  - AssetId:
-      m_storage: 9baa0140-ff74-421e-affd-efca8cebec4b
-    Asset: {fileID: 21300000, guid: 01e02995805eb483690380a911a657e2, type: 3}
-  - AssetId:
-      m_storage: 14e1ad32-9213-4da4-8313-71f0ffed36e6
-    Asset: {fileID: 21300000, guid: 8e8fc4c5d417b754280cd920d79a5e7f, type: 3}
-  - AssetId:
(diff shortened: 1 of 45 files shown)


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProBroadcasterBase.cs
```diff
@@ -59,6 +59,8 @@ protected override TextMeshProBroadcasterChangeType CalculateDeltaChanges()
 
         protected override void SendCompleteChanges(IEnumerable<SocketEndpoint> endpoints)
         {
+            previousText = this.textMesh.text;
+            previousProperties = new TextMeshProperties(textMesh);
```
chrisfromwork: There was some weird behavior here where font details were sent prior to populating on this first call to SendCompleteChanges

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
```diff
@@ -1,11 +1,75 @@
...
+            {
+                return false;
+            }
+
+            return this == assetId;
+        }
+
+        public override int GetHashCode()
+        {
+            return FileIdentifier.GetHashCode() ^ Guid.ToString().GetHashCode();
```
chrisfromwork: @matthejo should we just default to FileIdentifier here compared to obtaining a hash code?

samples/Build2019Demo.Unity/Assets/Generated.StateSynchronization.AssetCaches/Resources/SpriteAssetCache.asset
```diff
@@ -14,176 +14,297 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   assets:
   - AssetId:
-      m_storage: 74fae14d-39c0-497e-ae6d-72f1f2084231
-    Asset: {fileID: 21300000, guid: d651fc00b471cfe44834a222e9433461, type: 3}
+      guid:
+        m_storage: 00000000-0000-0000-f000-000000000000
```
wiwei: >00000000-0000-0000-f000-000000000000 [](start = 19, length = 36)

This GUID seems to be strannnnnnnnnnngely full of zeros - is it just the case that some portion of our deterministic guids are going to look like this?

It's just interesting seeing a bunch of other GUIDs that look like regular GUIDs, and then seeing this guid (and the c000 one) which looks like a bunch of zeroes that then got smashed by a cosmic ray.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -272,16 +273,23 @@ public override void UpdateAssetCache()
...
+                    else
+                    {
+                        Debug.LogError($"Unable to obtain id for Asset: {asset.ToString()}");
+                    }
                 }
             }
 
             CleanUpUnused(oldAssets, unvisitedAssets);
-            assets = oldAssets.Values.ToArray();
+            assets = oldAssets.Values.OrderBy(x => x.AssetId.FileIdentifier).ThenBy(x => x.AssetId.Guid).ToArray();
```
wiwei: >OrderBy(x => x.AssetId.FileIdentifier).ThenBy(x => x.AssetId.Guid) [](start = 38, length = 66)

If we're changing the mechanics of how our assets are enumerated (i.e. the EnumerateAllAssets). it might be worth documenting in the surfaces where this is exposed (i.e. EnumerateAllAssets will enumerate assets on File Identifier and then Guid)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
```diff
@@ -1,11 +1,75 @@
...
+    internal class AssetId
+    {
+        public static AssetId Empty { get; } = new AssetId(System.Guid.Empty, -1);
+
+        public StringGuid Guid => guid;
+
+        [SerializeField]
+        private StringGuid guid;
+
+        public long FileIdentifier => fileIdentifier;
```
wiwei: >FileIdentifier [](start = 20, length = 14)

Can you add a comment about what "FileIdentifier" is?

It's possible that this location isn't the right place for the comment (i.e. it's somewhat abstract at this location) but I'm trying to get a little more clarity for a reader who sees an abstract number (i.e. file ID), then look at this code, and ideally there should be some explanation/direction that this point that guides them to what's up

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
```diff
@@ -1,11 +1,75 @@
...
+
+        public override bool Equals(object obj)
+        {
+            AssetId assetId = obj as AssetId;
+            if (assetId == null)
+            {
+                return false;
+            }
+
+            return this == assetId;
```
wiwei: >this == assetId; [](start = 19, length = 16)

Can you also leave a comment on why fileIdentifier isn't necessary for equality? Or would it be reasonable to have a debug assert here if there doesn't need to be a runtime check?

samples/Build2019Demo.Unity/Assets/Generated.StateSynchronization.AssetCaches/Resources/SpriteAssetCache.asset
chrisfromwork: I have seen guids like this for Unity's default assets. Things like cube and cone meshes have similar guids.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
wiwei: Ah, chatted offline, I misread this, please ignore :)

---
In reply to: [343786783](https://github.com/microsoft/MixedReality-SpectatorView/pull/264#discussion_r343786783) [](ancestors = 343786783)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
wiwei: >return FileIdentifier.GetHashCode() ^ Guid.ToString().GetHashCode(); [](start = 12, length = 68)

[https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=dotnet-plat-ext-3.0](https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=dotnet-plat-ext-3.0)

wiwei:
>             return new StringGuid { m_storage = rhs.ToString("D") };

Now that I'm seeing this in two places, what's the significance of "D"? Is it just a sentinel string (i.e. "private static string INVALID_GUID = "D"?)

---
Refers to: src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StringGuid.cs:17 in 07dadf2. [](commit_id = 07dadf2cb6069a82ed0135802985bd250d36c164, deletion_comment = False)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
chrisfromwork: It's used in the operator definition

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
chrisfromwork: going to keep this hash code function for now based on non-deterministic scenarios for that combine/dependency on .net extensions.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
andreiborodin: Wouldn't we want to orderby Guid first, then file Id? As Guid represents the file, and fileId is the id within that file

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
```diff
@@ -1,11 +1,81 @@
...
+        }
+
+        public override int GetHashCode()
+        {
+            return FileIdentifier.GetHashCode() ^ Guid.ToString().GetHashCode();
```
andreiborodin: This will do an allocation everytime you add to the dictionary, because GetHashCode is called, which will create a string from Guid.

Instead, just do Guid.GetHashCode().

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
```diff
@@ -1,11 +1,81 @@
...
+        }
+
+        public static bool operator ==(AssetId lhs, AssetId rhs)
+        {
```
andreiborodin: I would simplify this to: Equals(lhs, rhs); which would propogate to your Equals override above.

I would update that method to not use this one, as you already did a null check, and "this" can't be null. So just compare Guid and FileIndentierifer.

Also, don't do Guid.ToString() for comparison, just compare Guids. Guid.Tostring() is an allocation

allotropeijk: Sounds good. I'll pick these changes up in stabilization/1.1.0 with the expectation to merge back into master

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
chrisfromwork: Sounds good. I'll pick these changes up in stabilization/1.1.0 with the expectation to merge back into master

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
chrisfromwork: hm, what aspect of Guid.ToString() is an allocation? StringGuid has a custom ToString definition that just returns the internally stored string value for the Guid. I'm not 100% if historically this type was made for serializtion, but we aren't currently using the default Guid type.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
```diff
@@ -1,11 +1,81 @@
...
+            {
+                return true;
+            }
+            else if ((object)lhs == null && (object)rhs != null ||
+                (object)lhs != null && (object)rhs == null)
+            {
+                return false;
+            }
+
+            return (lhs.Guid.ToString() == rhs.Guid.ToString()) && (lhs.FileIdentifier == rhs.FileIdentifier);
```
ivte-ms: >Guid [](start = 24, length = 4)

.Equals() please.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StringGuid.cs
```diff
@@ -7,7 +7,7 @@
 namespace Microsoft.MixedReality.SpectatorView
 {
     [Serializable]
-    internal class StringGuid
+    internal class StringGuid : IComparable
```
ivte-ms: >StringGuid [](start = 19, length = 10)

Out of curiosity, what's wrong with the ordinary GUID? It's already comparable, and should be way faster to work with.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCacheEntry.cs
ivte-ms:
```
        public override string ToString()
        {
            return (m_storage == null) ? System.Guid.Empty.ToString("D") : m_storage;
        }
```
Now looking at:
https://github.com/microsoft/referencesource/blob/master/mscorlib/system/guid.cs

```
            if (formatCh == 'D' || formatCh == 'd') {
                guidString = string.FastAllocateString(36);
            }
```
Looks like there is no special-casing for an empty guid, unless I missed it.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StringGuid.cs
chrisfromwork: I'm not sure historically why the GUID wasn't used. is a GUID serializable?

ivte-ms: The docs say that it [has](https://docs.microsoft.com/en-us/dotnet/api/system.guid?view=netframework-4.8) SerializableAttribute, so it should be.

I'm genuinely curious why the string representation was chosen instead. Maybe there was some issue with the Unity integration?

PR #267, opened 2019-11-05
Abstract Spectator View's networking components

chrisfromwork:
This change addresses: https://github.com/microsoft/MixedReality-SpectatorView/issues/186

Changes:
1) INetworkConnection and INetworkConnectionManager have been introduced to abstract out the SocketEndpoint and TCPConnectionManager components.
2) Spectator View classes previously referencing TCPConnectionManager that did not implement INetworkManager now reference INetworkManager for networking logic.
3) INetworkManager components (StateSynchronizationBroadcaster, StateSynchronizationObserver, etc) now create INetworkConnectionManagers based on prefabs. A default prefab can be specificed for each component, NetworkConfigurationSettings also allows providing a component override.
4) INetworkConnectionManager (previously TCPConnectionManager) now exposes all of its known connections to components. This will allow components to broadcast their states when network connections occur prior to network connection event subscriptions.
5) Configuring the reconnection logic for sockets is now defined on the TCPConnectionManager in the TCPConnectionManager prefab. If folks opt to not use custom networking ui, they may need to update the property in this prefab.

Additional thoughts:
1) INetworkConnectionManager contains functions StartListening and ConnectTo. These functions feel specific to the functionality provided in TCPConnectionManager compared to generally reusable functionality. It's a bit unclear how this interface will be used in the future. We should wait for user feedback to understand how this could/should be modified.

### Notes:
This review updates a lot of locations to use INetworkConnection that previously relied on SocketEndpoint class definitions. This was needed so that different endpoint/connections could be used in place of SocketEndpoints. Now, developers will be able to use their own networking stack if they so desire.

### Migration Instructions:
1. Update any custom usage of SocketEndpoint to use INetworkConnection methods.
1. Update any custom network connection managers (that were used in place of TCPConnectionManager) to implement INetworkConnectionManager and place them in prefabs to be instantiated by various INetworkManager components in the Spectator View codebase.


src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CalibrationDataProvider.cs
```diff
@@ -18,23 +18,27 @@ namespace Microsoft.MixedReality.SpectatorView
     /// Loads calibration data from the Pictures library on the device and transfers that data
     /// to the compositor upon connection.
     /// </summary>
-    [RequireComponent(typeof(TCPConnectionManager))]
+    [RequireComponent(typeof(INetworkManager))]
```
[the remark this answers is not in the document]
chrisfromwork: no, this doesn't create errors in the editor similar to class types. I'm not sure what preference we have here, I would like to indicate to developers that you need an INetworkManager when using this component. RequireComponent dynamicaly creates components when they don't exist, so you can't declare something like NetworkManager or NetworkManager<T> as the required type. We also don't want these components to require a specific NetworkManager implementation like StateSynchronizationBroadcaster or StateSynchronizationObserver

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Socketer/TCPSocketEndpoint.cs
```diff
@@ -7,17 +7,15 @@
 
 namespace Microsoft.MixedReality.SpectatorView
 {
-    public sealed class SocketEndpoint
+    public sealed class TCPSocketEndpoint : INetworkConnection
```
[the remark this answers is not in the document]
chrisfromwork: @ivte-ms do you have  preference on endpoint vs socket? Or really anyone on this review? :)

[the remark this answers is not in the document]
stephenatwork: Looking at this again. Given that we have:
```
INetworkConnectionManager
INetworkConnection
```
I propose `SocketEndpoint` -> `TCPNetworkConnection`

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CalibrationDataProvider.cs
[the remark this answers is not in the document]
stephenatwork: If this is only for documentation and doesn't have any effect, then we're good.
What about this advice?
https://answers.unity.com/questions/222580/requirecomponent-but-dont-automatically-add-it.html
```
 void OnEnable() {
     if (GetComponent<INetworkManager>() == null) {
         throw new MissingComponentException("Missing network manager component");
     }
 }
```

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/ComponentBroadcaster.cs
```diff
@@ -182,56 +182,56 @@ private void EnsureComponentInitialized()
...
-                    if (ShouldSendChanges(endpointsNeedingComponentCreation[i]))
+                    if (ShouldSendChanges(connectionsNeedingComponentCreation[i]))
                     {
-                        SendComponentCreation(endpointsNeedingComponentCreation[i]);
-                        endpointsNeedingComponentCreation.RemoveAt(i);
+                        SendComponentCreation(connectionsNeedingComponentCreation[i]);
+                        connectionsNeedingComponentCreation.RemoveAt(i);
                     }
                 }
             }
```
ivte-ms: Can we make it a bit more idiomatic? Something like:
`connectionsNeedingComponentCreation?.RemoveAll(x => SendComponentCreationIfShould(x));`
where `bool SendComponentCreationIfShould(...)` is also directly reusable below.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/ComponentBroadcaster.cs
```diff
@@ -182,56 +182,56 @@ private void EnsureComponentInitialized()
...
                 else
                 {
-                    if (endpointsNeedingComponentCreation == null)
+                    if (connectionsNeedingComponentCreation == null)
                     {
-                        endpointsNeedingComponentCreation = new List<SocketEndpoint>();
+                        connectionsNeedingComponentCreation = new List<INetworkConnection>();
                     }
-                    endpointsNeedingComponentCreation.Add(newConnection);
+                    connectionsNeedingComponentCreation.Add(newConnection);
```
ivte-ms:
```
(connectionsNeedingComponentCreation ??= new List<INetworkConnection>())
    .Add(newConnection);
```

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
```diff
@@ -142,27 +146,33 @@ protected override void OnDestroy()
                 connectionManager.OnReceive -= OnReceive;
             }
 
+            if (connectionManagerGameObject != null)
+            {
+                connectionManager = null;
```
ivte-ms: Why are we doing it conditionally here (as opposed, for example, the block above)?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Socketer/TCPConnectionManager.cs
```diff
@@ -3,61 +3,75 @@
...
+        {
+            get
+            {
+                List<INetworkConnection> connections = new List<INetworkConnection>();
+                if (clientConnection != null)
+                {
+                    connections.Add(clientConnection);
+                }
+
+                if (serverConnections.Count > 0)
```
ivte-ms: `if (!serverConnections.IsEmpty())` please.

The internal implementation of `ConcurrentDictionary` [works like this](https://github.com/microsoft/referencesource/blob/master/mscorlib/system/collections/Concurrent/ConcurrentDictionary.cs):

- `IsEmpty()` returns immediately as soon as it finds any elements in any of the tables.
- `Count` won't stop until it iterates over **all of them**, adding all results together, even if it was obvious after the first table that there are elements.

Fixing other 12 instances of "Count > 0" would also be great. Don't know how many of them will become faster, but IsEmpty is just more idiomatic anyway.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/ComponentBroadcaster.cs
stephenatwork: I wasn't sure if ??= was an encoding error or a misplaced C++ trigraph! Turns out it's C# 8.0. Would there be any gotchas with bumping the langversion?  https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationBroadcaster.cs
```diff
@@ -102,7 +87,7 @@ public bool HasConnections
         {
             get
             {
-                return connectionManager != null && connectionManager.HasConnections;
+                return connectionManager != null && connectionManager.Connections.Count > 0;
```
ivte-ms: If I'm reading this correctly, connectionManager.Connections allocates and populates a new list just so that we can tell that connections exist.
TCPConnectionManager already has a bit more reasonable HasConnections property (which can be further implemented, see the comment about Count there). It's probably worth declaring it in INetworkConnectionManager (and then using it here).

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/DynamicGameObjectHierarchy/DynamicGameObjectHierarchyBroadcaster.cs
```diff
@@ -85,30 +85,30 @@ protected override void ProcessNewConnections(IEnumerable<SocketEndpoint> connec
...
 
-        protected override void SendCompleteChanges(IEnumerable<SocketEndpoint> endpoints)
+        protected override void SendCompleteChanges(IEnumerable<INetworkConnection> connections)
         {
-            foreach (SocketEndpoint endpoint in endpoints)
+            foreach (INetworkConnection connection in connections)
             {
                 PerConnectionInstantiationState state;
-                if (!perConnectionInstantiationState.TryGetValue(endpoint, out state))
+                if (!perConnectionInstantiationState.TryGetValue(connection, out state))
```
ivte-ms: Not sure what's going on here. Either
`if (!perConnectionInstantiationState.ContainsKey(connection))`
or there is a bug and `state` was supposed to be used somehow.

See similar usages of TryGetValue

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/ComponentBroadcaster.cs
ivte-ms: If it's not already supported by Unity, then that's probably outside of our control.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Socketer/TCPSocketEndpoint.cs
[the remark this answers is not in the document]
ivte-ms: Just by Wiki's definition, all [sockets are endpoints](https://en.wikipedia.org/wiki/Network_socket), but not all endpoints are sockets. Also, neither is necessarily a connection (there is a paragraph about that).

Personally, I'm strongly against adding additional levels of confusion, and would prefer if all sending-related methods would operate on very abstract endpoints when possible, that have as few methods beyond Send() as possible (they may or may not represent a single connection, and may or may not be implemented with sockets; that's an implementation detail), but I remember @andreiborodin wasn't happy with the word "Endpoint" at some point in the past.

The layer that transports messages for StateSync will be using such lightweight abstractions. But if you feel that here we'll benefit from heavier and more specific interfaces, I'm fine with that. Adapting one to another shouldn't be difficult.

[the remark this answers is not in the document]
stephenatwork: I agree that endpoints & send() are a worthy goal. As it stands though, SpecatorView has a few places which refer to connections, not abstract endpoints. i.e. INetworkConnection.IsConnected and INetworkConnection.Disconnect(), INetworkManager.Connections + more.
So for the code as it currently stands, connection seems more apt.

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CalibrationDataProvider.cs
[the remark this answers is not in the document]
chrisfromwork: sounds good to me

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
```diff
@@ -2,35 +2,36 @@
...
 using System.IO;
 using UnityEngine;
 
 namespace Microsoft.MixedReality.SpectatorView
 {
     public abstract class NetworkManager<TService> : CommandRegistry<TService>, INetworkManager where TService : Singleton<TService>
     {
-        private float lastReceivedUpdate;
-
+#pragma warning disable 414 // The field is assigned but its value is never used
```
fibann: It seems that this _is_ used below. What's the deal?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Networking/INetworkConnection.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Returns true if the network connection is connected, otherwise false
+        /// </summary>
+        bool IsConnected { get; }
+
+        /// <summary>
+        /// Call to set the network connection state to disconnected
+        /// </summary>
+        void Disconnect();
```
ivte-ms: Not entirely clear what's the interaction between this and QueueIncomingMessages()

Consider the scenario:

- I received a handshake message.
- I checked some security token.
- It didn't match, so I immediately closed the connection.
- Now I expect that I'll never receive a message from this connection after I closed it, and therefore my message processing code can assume that the token is already verified, and have no checks about that whatsoever.

Do provide a guarantee like this, or some in-flight zombie messages can also be processed afterwards? (the question may be meaningless depending on what QueueIncomingMessages() actually does, please ignore if it doesn't apply)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
```diff
@@ -2,35 +2,36 @@
...
+        [Tooltip("Default prefab for creating an INetworkConnectionManager.")]
         [SerializeField]
-        protected TCPConnectionManager connectionManager = null;
+        private GameObject defaultConnectionManagerPrefab = null;
+        private GameObject connectionManagerGameObject = null;
+        protected INetworkConnectionManager connectionManager = null;
+#pragma warning restore 414
 
-        public bool AttemptReconnectWhenClient
```
fibann: The reconnection logic across a few files seems to have been moved (removed?) but this is not mentioned in the PR description. What happened to it?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationObserver.cs
```diff
@@ -50,18 +50,7 @@ protected override void Awake()
...
-
-                // Start listening to incoming connections as well.
-                connectionManager.StartListening(port);
-            }
-            else
-            {
-                Debug.LogError("Connection manager not specified for Observer.");
-            }
-
+            StartListening(port);
```
fibann: IIRC this listening socket is unused. Can we remove it since we are touching this code?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Socketer/TCPConnectionManager.cs
chrisfromwork: it looks like this function exists for ConcurrentDictionary but doesnt exist for Dictionary. I will update ConcurrentDictionaries.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Socketer/TCPSocketEndpoint.cs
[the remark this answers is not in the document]
chrisfromwork: I think we can migrate to these improvements over time. Everything other than Send seems like an implementation specific detail remaining from the TCPConnectionManager. Ideally, folks will provide feedback when attempting to integrate their own networking stacks and we can use that feedback to reduce/improve this interface.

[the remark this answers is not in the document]
chrisfromwork: I will change SocketEndpoint to TCPNetworkConnection for now

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
chrisfromwork: stale code, removing now.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
chrisfromwork: good call out.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/DynamicGameObjectHierarchy/DynamicGameObjectHierarchyBroadcaster.cs
chrisfromwork: i think this is just old code, i didn't touch this implementation but will move towards ContainsKey

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Socketer/TCPNetworkConnection.cs
```diff
@@ -96,15 +82,12 @@ public void StopConnectionAttempts()
...
-        /// Call to send data to this endpoint
-        /// </summary>
-        /// <param name="data">data to send</param>
+        /// <inheritdoc />
         public void Send(byte[] data)
         {
             if (!IsConnected)
             {
-                Debug.LogWarning("Attempted to send message to disconnected SocketEndpoint.");
+                Debug.LogWarning("Attempted to send message to disconnected TCPNetworkConnection.");
```
matthejo: >TCPNetworkConnection [](start = 76, length = 20)

Could use string interpolation with {nameof(TCPNetworkConnection)} here for future refactoring

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/ComponentBroadcaster.cs
chrisfromwork: keeping this this way for now. This would be a good suggestion in the future.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemManager.cs
```diff
@@ -523,6 +523,15 @@ private void RegisterEvents(INetworkManager networkManager)
...
             networkManager.RegisterCommandHandler(SupportedLocalizersMessageHeader, OnSupportedLocalizersMessageReceived);
             networkManager.RegisterCommandHandler(SpatialCoordinateSystemParticipant.LocalizationDataExchangeCommand, OnParticipantDataReceived);
+            
+            if (networkManager.IsConnected)
+            {
+                var connections = networkManager.Connections;
+                foreach (var connection in connections)
+                {
+                    OnConnected(connection);
+                }
```
matthejo: If OnConnected caused the connection status to change (e.g. a disconnect), this could potentially cause enumeration over a modified collection. Might be safer to create a snapshot copy of the connections to visit.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/ComponentBroadcaster.cs
chrisfromwork: Opted for TrySendComponentCreation

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
```diff
@@ -2,30 +2,30 @@
...
-        {
-            get { return attemptReconnectWhenClient; }
-            set { attemptReconnectWhenClient = value; }
-        }
-        private bool attemptReconnectWhenClient = true;
+        private float lastReceivedUpdate;
+        private INetworkConnection currentConnection;
 
-        private SocketEndpoint currentConnection;
+        private static readonly List<INetworkConnection> emptyConnectionList = new List<INetworkConnection>();
```
matthejo: >private static readonly List<INetworkConnection> emptyConnectionList = new List<INetworkConnection>(); [](start = 8, length = 102)

This could be "private static readonly IReadOnlyList<INetworkConnection> emptyConnectionList = Array.Empty<INetworkConnection>();"  An empty array implements IReadOnlyList and requires less space in memory (and is truly read-only).

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
```diff
@@ -2,30 +2,30 @@
...
 
         /// <inheritdoc />
         public string ConnectedIPAddress => currentConnection?.Address;
 
+        public IReadOnlyList<INetworkConnection> Connections => connectionManager == null ? emptyConnectionList : connectionManager.Connections;
```
matthejo: > public IReadOnlyList<INetworkConnection> Connections => connectionManager == null ? emptyConnectionList : connectionManager.Connections; [](start = 7, length = 137)

Missing doc comment

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
chrisfromwork: oh cool, this is exactly what i was looking for :)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
```diff
@@ -174,7 +165,33 @@ protected void OnReceive(IncomingMessage data)
...
+                var prefab = defaultConnectionManagerPrefab;
+                if (NetworkConfigurationSettings.IsInitialized &&
+                    NetworkConfigurationSettings.Instance.OverrideConnectionManagerPrefab != null)
+                {
+                    prefab = NetworkConfigurationSettings.Instance.OverrideConnectionManagerPrefab;
+                }
+
+                if (prefab == null)
+                {
+                    throw new MissingComponentException("Network connection manager prefab wasn't specified. DeviceInfoBroadcaster will not work correctly.");
```
matthejo: >DeviceInfoBroadcaster [](start = 109, length = 21)

DeviceInfoBroadcaster looks like it might have been a copy-pasted type name from somewhere else since this is on NetworkManager

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationObserver.cs
chrisfromwork: i would prefer to keep this for now, i'm not sure the impact removing this will have on all video camera/editor connection items.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
chrisfromwork: hm, i felt like this reconnect logic was specific to the TCPConnectionManager compared to the NetworkManager. We could add this to the INetworkConnectionManager interface definition, but i don't know that all network connection managers would handle this functionality.

Right now if you want to disable attempting reconnections or enable, you need to update/create a new TCPConnectionManager prefab. I'll add this to the review description.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Networking/INetworkConnection.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Checks whether the associated client is still active. If not, the client is disconnected.
+        /// </summary>
+        /// <param name="currentTime">Time to use relative to last active timestamp to determine whether to disconnect</param>
+        void CheckConnectionTimeout(DateTime currentTime);
```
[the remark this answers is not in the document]
chrisfromwork: we don't ever use this funcitonality. I am going to remove it for now

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Networking/INetworkConnection.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Microsoft.MixedReality.SpectatorView
+{
+    /// <summary>
+    /// Classes that implement this interface represent network connections and allow sending messages across a network.
+    /// </summary>
+    public interface INetworkConnection
+    {
+        /// <summary>
+        /// IP address for the connected peer
```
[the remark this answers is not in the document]
chrisfromwork: I think I would prefer this to be a ToString method compared to Address. We use this most frequently for displaying connection information.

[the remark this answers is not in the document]
chrisfromwork: If addresses are ever actually needed, we can add this requirement then.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationObserver.cs
chrisfromwork: @matthejo  any thoughts on whether or not this socket is used?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
chrisfromwork: Updated the description.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Networking/INetworkConnection.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// Checks whether the associated client is still active. If not, the client is disconnected.
+        /// </summary>
+        /// <param name="currentTime">Time to use relative to last active timestamp to determine whether to disconnect</param>
+        void CheckConnectionTimeout(DateTime currentTime);
+
+        /// <summary>
+        /// Call to start enqueuing incoming messages
+        /// </summary>
+        /// <param name="incomingQueue">Queue used for enqueuing messages</param>
+        void QueueIncomingMessages(ConcurrentQueue<IncomingMessage> incomingQueue);
```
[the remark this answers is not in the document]
chrisfromwork: this method reads in a bit of an unclear manner, but basically it sets up a queue so that you can start receiving messages from the connection. It's a bit unclear what happens if multiple queues are provided. For now, I think we should do a TrySubscribeToIncomingMessages and return false if a subscription already exists. Only our TCPConnectionManager should be providing a queue currently. overtime we can change this as needed but having each component subscribe to viewing this data seems like bad practice.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Networking/INetworkConnection.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Call to stop enqueuing incoming messages
+        /// </summary>
+        void StopIncomingMessageQueue();
```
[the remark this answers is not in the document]
chrisfromwork: Moving this to UnsubscribeToIncomingMessages. If folks would pref this to be a TryUnsubscribeToIncomingMessages I can update implementation, but I think this should be viewed as a subscription/unsubscription to messages, or at least that is how it is used by TCPConnectionManager.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Networking/INetworkConnection.cs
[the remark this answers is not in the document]
ivte-ms: May I suggest something like SetMessageQueue() then?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Networking/INetworkConnection.cs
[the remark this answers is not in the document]
ivte-ms: Suggestion: SetMessageQueue(null) (see the comment above) removes this method.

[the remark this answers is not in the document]
chrisfromwork: SetMessageQueue(null) doesn't have any safe guard to make sure that you aren't unsetting someone else's message queue. In this context, we only have we only have one setter, but this interface is public so folks may try and use it in a different manner than we do currently. I'd prefer having this as TrySubscribe/Unsubscribe so that we don't rule out multiple listeners. I'd also prefer not to rely SetMessageQueue(null) as a way of unregistering in case multiple users interact with this interface.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Networking/INetworkConnection.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Call to send data to the connected peer
+        /// </summary>
+        /// <param name="data">data to send</param>
+        void Send(byte[] data);
```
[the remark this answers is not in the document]
chrisfromwork: The majority of our send calls take arrays that aren't reused by the sender. It should be straight forward to adopt either of these preferences.

TCPNetworkConnection wraps different socket implementations based on platform/use cases. In some cases, we appear to immediately call System.NET.Sockets.Send, which blocks/throws. In other cases we use System.Net.Sockets.TcpClient and Windows.Networking.Sockets.StreamSocket to create and write to streams.

I would prefer moving towards something like ReadOnlySpan like you've called out. It looks like said type might not exist for the .NET version we are using.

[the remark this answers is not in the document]
stephenatwork: I think the poor mans ReadOnlySpan is (byte[], int offset, int len). I've seen that signature a lot in the pre-span interfaces. The byte array is borrowed for the duration of the call so the caller can often save a copy.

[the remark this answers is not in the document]
ivte-ms: @stephenatwork (byte[], int offset, int len) is not making it clear that the implementation is not passing all three args to some async without making a hard copy.

My personal vote is to have two independent calls, as described above. This way both the caller and the interface can do the best thing possible.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/INetworkManager.cs
```diff
@@ -54,7 +59,7 @@ public interface INetworkManager : ICommandRegistry
         /// <summary>
         /// Send a packet of data to all connected devices.
         /// </summary>
-        /// <param name="data">The data to send to each connected device.</param>
-        void Broadcast(byte[] data);
+        /// <param name="data">The data to send to each connected device. The reference is set to null once taking over ownership</param>
+        void Broadcast(ref byte[] data);
```
stephenatwork: This reads strangely - it basically forces the caller to make a copy.
A familiar idiom which avoids the copy is:
```
void Broadcast(byte[] data, int offset=0, int len=-1); // if len==-1 data.len is used instead
```
Then, callers would look like this
```
using (MemoryStream stream = new MemoryStream())
using (BinaryWriter message = new BinaryWriter(memoryStream)) {
                message.Write(...);
                connectionManager.Broadcast(stream.GetBuffer(), 0, stream.Position);
}

Comment (stephenatwork) [reply src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/INetworkManager.cs]: I looked a bit more and almost all of the callers of Broadcast are creating their own local stream & binarywriter, writing a few bytes and discarding them both. Another alternative is to put the stream & writer in the manager like this:
```
void Broadcast( Action<BinaryWriter> action );
```
Here's a sample of some of the current code (SetPerformanceMonitoringMode)
```
                byte[] message;
                using (MemoryStream stream = new MemoryStream())
                using (BinaryWriter writer = new BinaryWriter(stream))
                {
                    writer.Write(PerfDiagnosticModeEnabledCommand);
                    writer.Write(enabled);
                    writer.Flush();
                    message = stream.ToArray();
                }
                connectionManager.Broadcast(message);
```
With the action idiom, it it more concise, no copying & no garbage created.
```
                connectionManager.Broadcast( (writer) => {
                    writer.Write(PerfDiagnosticModeEnabledCommand);
                    writer.Write(enabled);
                    writer.Flush();
                });
```

stephenatwork: On second thoughts, I'm not sure if creation of that lambda produces garbage. Any C# experts know?

andreiborodin: As I understand lambda causes an allocation if it captures a local variable. 

When it contains a set of self-contained statements, it gets compiled to a static method at compile time that is referenced through a Delegate. 

If it references the values of current instance class, then, it gets compiled to the instance method of the class and passed through a delegate.

If it captures a local variable, then a class is created with a member for that variable and the instance method. That variable is assigned, and the instance method is passed through a Delegate.

andreiborodin: Your example will result in a static method at compile time, and no allocation.

stephenatwork: Thanks Andrei! I read a bit more and it looks like there are 2 parts to this.
1. the compiler may create an object for the captures or use a static or class method.
2. the compiler creates an instance of Action<T>
So while we're can avoid \#1, exactly as you say, it seems there's no way to avoid \#2. I verified this with ildasm :/
Still, there are fewer allocations and virtual calls (e.g. we save 2 Disposes)

chrisfromwork: I will address this in a follow up review

PR #269, opened 2019-11-06
Fix text mesh text updates on android

chrisfromwork:
This review addresses: https://github.com/microsoft/MixedReality-SpectatorView/issues/263

TextMeshPro content synchronization seems to be working correctly, but it appears that the first declaration of fonts for the TextMeshPro item doesn't result in the component being updated correctly. In this change we wait an update to force a text mesh component update, which gets things to appear correctly.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Diagnostics;
 using System.IO;
 
 #if STATESYNC_TEXTMESHPRO
@@ -13,6 +14,8 @@ namespace Microsoft.MixedReality.SpectatorView
 {
     internal abstract class TextMeshProObserverBase : ComponentObserver
     {
+        private bool needsUpdate = false;
+
 #if STATESYNC_TEXTMESHPRO
         protected TMP_Text TextMeshObserver
         {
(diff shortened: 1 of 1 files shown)


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
```diff
@@ -119,8 +121,19 @@ public override void Read(SocketEndpoint sendingEndpoint, BinaryReader message)
                 TextMeshObserver.verticalMapping = (TextureMappingOptions)message.ReadByte();
                 TextMeshObserver.wordWrappingRatios = message.ReadSingle();
                 TextMeshObserver.wordSpacing = message.ReadSingle();
+
+                needsUpdate = true;
```
ivte-ms: When is this method called, relative to Update, and why can't we do the ForceMeshUpdate() right here?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
```diff
@@ -119,8 +121,19 @@ public override void Read(SocketEndpoint sendingEndpoint, BinaryReader message)
...
             }
 #endif
         }
+
+        protected virtual void Update()
+        {
+            if (needsUpdate)
+            {
+                TextMeshObserver.ForceMeshUpdate();
```
matthejo: >TextMeshObserver.ForceMeshUpdate(); [](start = 16, length = 35)

I think it would be good to put a comment explaining why this is necessary

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
chrisfromwork: For whatever reason it doesn't resolve the solution. We have also tried changing to the execution order for component observers, but attempting to force an update on the same update pass the text mesh pro text is created doesnt result in the font resolving correctly/things rendering correctly.

ivte-ms: Hm. It's just... if that's something that happens after the update pass, we won't get the result before the next frame anyway. In that case delaying it with Invoke() could have a benefit that the class won't have an Update() method (so the reflection won't pick it up, so it won't be participating in the update pass every frame, even if there is nothing to update). But that's probably not a big deal.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
chrisfromwork: sounds good

PR #282, opened 2019-11-21
Add unsupported localizer list

chrisfromwork:
Some changes in the current release enabled new localizers in the editor for simulation purposes. This makes for a confusing experience in the compositor window because it leads to unsupported localizers being displayed for video camera localization. This change adds an unsupported localizer list that prevents these unsupported video camera localizers from being displayed as options. In the future, we can add an extension point here for customization if customer asks arise.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindowBase.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindowBase.cs
@@ -19,6 +19,15 @@ namespace Microsoft.MixedReality.SpectatorView.Editor
         protected const string AppDeviceTypeLabel = "App";
         protected string holographicCameraIPAddress;
 
+        private readonly List<Guid> unsupportedLocalizers = new List<Guid>
+        {
+            ArUcoMarkerVisualSpatialLocalizer.Id,
+            ArUcoMarkerVisualDetectorSpatialLocalizer.Id,
+            QRCodeMarkerVisualSpatialLocalizer.Id,
+            QRCodeMarkerVisualDetectorSpatialLocalizer.Id,
+            new Guid("FB077E49-B855-453F-9FB5-77187B1AF784") // SpatialAnchorsLocalizer (There is no asmdef to reference for ASA)
+        };
+
         private CompositionManager cachedCompositionManager;
         private HolographicCameraObserver cachedHolographicCameraObserver;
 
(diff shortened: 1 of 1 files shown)


src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindowBase.cs
```diff
@@ -19,6 +19,15 @@ internal class CompositorWindowBase<TWindow> : EditorWindowBase<TWindow> where T
         protected const string AppDeviceTypeLabel = "App";
         protected string holographicCameraIPAddress;
 
+        private readonly List<Guid> unsupportedLocalizers = new List<Guid>
```
wiwei: >private readonly List<Guid> unsupportedLocalizers = new List<Guid> [](start = 8, length = 66)

Can we also have a comment here that explains more about the unsupported nature of this list? i.e. down below where it's used, we join two different list - both an unsupported and a supported one. A cursory reading leaves a reader somewhat confused because it seems like everything that is supported... should probably be all you need to check.

The other possibility is to rename things down below (i.e. not "supported" but "candidate" localizers).

Either way, having some more context about why this additional filtering is helpful.

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindowBase.cs
```diff
@@ -19,6 +19,18 @@ internal class CompositorWindowBase<TWindow> : EditorWindowBase<TWindow> where T
         protected const string AppDeviceTypeLabel = "App";
         protected string holographicCameraIPAddress;
 
+        /// <summary>
+        /// This list contains localizers that can't be used in the compositor window for video camera filming.
+        /// </summary>
+        private readonly List<Guid> unsupportedLocalizers = new List<Guid>
```
matthejo: Would it potentially be better to have a list of "compositorSupported" localizers, and then list the ones that are supported? It seems to me like any new future localizers are likely to not be supported by default unless something is done to them to make them work.

chrisfromwork: I think I would prefer to have an exclusion list. compositorSupported localizers would require us to add some mechanism for registering localizers as supported for external contributors. Or they would have to edit this source which may be distributed through nuget as compiled binaries.

PR #283, opened 2019-11-22
Fix a warning of the unused needsUpdate field.

wiwei:
When importing a fresh SpectatorView pull, you'll see this warning in the Unity console because this field is only used when STATESYNC_TEXTMESHPRO is specified. Of course when you're opening things for the first time (i.e. ingesting SV for the first time) it's possible to have made this step (because adding that define is a step much later)

This just puts it guarded by the define that it's actually used in.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
@@ -13,9 +13,9 @@ namespace Microsoft.MixedReality.SpectatorView
 {
     internal abstract class TextMeshProObserverBase : ComponentObserver
     {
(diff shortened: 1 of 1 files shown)


chrisfromwork:
Retargetting the stabilization/1.1.0 branch, we will merge this back into master after getting out a release.

Did you mean to include/update the MRTK submodule? Was this something that automatically happened when running the setup scripts?

wiwei:
> Did you mean to include/update the MRTK submodule? Was this something that automatically happened when running the setup scripts?

I'd gotten through the README to this point:

https://github.com/microsoft/MixedReality-SpectatorView#basic-unity-setup

I think that the instructions to add the flag are here:

https://github.com/microsoft/MixedReality-SpectatorView/blob/75361439f98e1d96208c9f5960bc3eac64381e97/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/README.md

Which I think makes the warning go away - but note that people probably have read it at that point, so they will see this warning.

chrisfromwork:
forgot to sign off, but I meant to approve this review :)

PR #286, opened 2019-11-25
Add tools for assessing summed times and memory usage to Spectator View

chrisfromwork:
We are looking to dynamically load asset bundles, which calls into question memory usage/performance around loading/using assets. This review adds some new items to the tooling for assessing performance information:

1) BroadcasterSettings now contains a flag that enables performance monitoring from Spectator View startup. This is needed to understand the time/memory associated with loading asset caches.
2) AssetCaches now load content on Start compared to on Awake. Loading on awake caused asset caches to load prior to BroadcasterSettings getting instantiated. This prevented seeing a flag that indicated whether to default to using performance reporting.
3) StateSynchronizationPerformanceMonitor now contains IncrementEventDuration. IncrementEventDuration allows developers to assess the total amount of time put towards specific events compared to assessing the average time associated with an event. This allows for viewing things like the total amount of time put towards asset loading.
4) StateSynchronizationPerformanceMonitor now contains MeasureEventMemoryUsage. This uses Unity's Performance monitor to assess changes in total allocated memory, etc over the duration of an event. This can give rough estimates for understanding how much memory is allocated when loading asset caches.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
@@ -21,7 +21,11 @@ namespace Microsoft.MixedReality.SpectatorView
         public static TAssetCache LoadAssetCache<TAssetCache>()
             where TAssetCache : AssetCache
         {
-            return Resources.Load<TAssetCache>(typeof(TAssetCache).Name);
+            using (StateSynchronizationPerformanceMonitor.Instance.IncrementEventDuration(typeof(TAssetCache).Name, "LoadAssetCache"))
+            using(StateSynchronizationPerformanceMonitor.Instance.MeasureEventMemoryUsage(typeof(TAssetCache).Name, "LoadAssetCache"))
+            {
+                return Resources.Load<TAssetCache>(typeof(TAssetCache).Name);
+            }
         }
 
         public static string GetAssetPath(string assetName, string assetExtension)
@@ -224,27 +228,35 @@ namespace Microsoft.MixedReality.SpectatorView
 
         public TAsset GetAsset(AssetId assetId)
         {
-            TAssetEntry assetEntry;
-            if (LookupByAssetId.TryGetValue(assetId, out assetEntry))
+            using (StateSynchronizationPerformanceMonitor.Instance.IncrementEventDuration(this.GetType().Name, "GetAsset"))
+            using (StateSynchronizationPerformanceMonitor.Instance.MeasureEventMemoryUsage(this.GetType().Name, "GetAsset"))
             {
-                return assetEntry.Asset;
-            }
-            else
-            {
-                return default(TAsset);
+                TAssetEntry assetEntry;
+                if (LookupByAssetId.TryGetValue(assetId, out assetEntry))
+                {
+                    return assetEntry.Asset;
+                }
+                else
+                {
+                    return default(TAsset);
+                }
             }
         }
 
         public AssetId GetAssetId(TAsset asset)
         {
-            TAssetEntry assetEntry;
-            if (asset != null && LookupByAsset.TryGetValue(asset, out assetEntry))
+            using (StateSynchronizationPerformanceMonitor.Instance.IncrementEventDuration(this.GetType().Name, "GetAssetId"))
+            using (StateSynchronizationPerformanceMonitor.Instance.MeasureEventMemoryUsage(this.GetType().Name, "GetAssetId"))
             {
-                return assetEntry.AssetId;
-            }
-            else
-            {
-                return AssetId.Empty;
+                TAssetEntry assetEntry;
+                if (asset != null && LookupByAsset.TryGetValue(asset, out assetEntry))
+                {
+                    return assetEntry.AssetId;
+                }
+                else
+                {
+                    return AssetId.Empty;
+                }
             }
         }
 
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
@@ -229,7 +229,6 @@ namespace Microsoft.MixedReality.SpectatorView
         public TAsset GetAsset(AssetId assetId)
         {
             using (StateSynchronizationPerformanceMonitor.Instance.IncrementEventDuration(this.GetType().Name, "GetAsset"))
-            using (StateSynchronizationPerformanceMonitor.Instance.MeasureEventMemoryUsage(this.GetType().Name, "GetAsset"))
             {
                 TAssetEntry assetEntry;
                 if (LookupByAssetId.TryGetValue(assetId, out assetEntry))
@@ -246,7 +245,6 @@ namespace Microsoft.MixedReality.SpectatorView
         public AssetId GetAssetId(TAsset asset)
         {
             using (StateSynchronizationPerformanceMonitor.Instance.IncrementEventDuration(this.GetType().Name, "GetAssetId"))
-            using (StateSynchronizationPerformanceMonitor.Instance.MeasureEventMemoryUsage(this.GetType().Name, "GetAssetId"))
             {
                 TAssetEntry assetEntry;
                 if (asset != null && LookupByAsset.TryGetValue(asset, out assetEntry))
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
@@ -22,7 +22,7 @@ namespace Microsoft.MixedReality.SpectatorView
             where TAssetCache : AssetCache
         {
             using (StateSynchronizationPerformanceMonitor.Instance.IncrementEventDuration(typeof(TAssetCache).Name, "LoadAssetCache"))
-            using(StateSynchronizationPerformanceMonitor.Instance.MeasureEventMemoryUsage(typeof(TAssetCache).Name, "LoadAssetCache"))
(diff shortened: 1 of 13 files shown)


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -21,7 +21,11 @@ internal abstract class AssetCache : ScriptableObject
         public static TAssetCache LoadAssetCache<TAssetCache>()
             where TAssetCache : AssetCache
         {
-            return Resources.Load<TAssetCache>(typeof(TAssetCache).Name);
+            using (StateSynchronizationPerformanceMonitor.Instance.IncrementEventDuration(typeof(TAssetCache).Name, "LoadAssetCache"))
```
stephenatwork: Will IncrementEventDuration and MeasureEventMemoryUsage be paired often? Looks like a dedicated method could be a perf win.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceMonitor.cs
```diff
@@ -28,22 +48,48 @@ public struct ParsedMessage
...
+        {
+            if (!StateSynchronizationPerformanceParameters.EnablePerformanceReporting)
+            {
+                return null;
+            }
+
+            var key = new PerformanceEventKey(componentName, eventName);
+            if (!incrementEventStopWatches.TryGetValue(key, out var stopwatch))
+            {
+                stopwatch = new Stopwatch();
```
stephenatwork: Can we avoid all these stopwatch objects? And therefore the map lookup also. Have a single static stopwatch and record the start/end ticks?

stephenatwork:
It's super handy to be able to leave profiling enabled almost all the time. So I started looking at possible performance wins to allow this. The local changes I started to suggest will only get us so far though.

It's a big change, but is there any way we can defer almost all the work until ReadMessage? I.e. IncrementEventCount would do just stash the parameters. (write(tag_inc); write(componentname); write(eventname))
MeasureEventDuration would do something similar( write(tag_duration_pop); write(....)) and the returned disposable would (write(tag_duration_pop)). Or if the disposable comes with a perf cost, have explicit push and pop methods.

Then all the map lookups and garbage producing operations could be deferred until the data is read.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceMonitor.cs
```diff
@@ -186,5 +295,42 @@ public void Dispose()
...
+                    memoryUsage == null)
+                {
+                    UnityEngine.Debug.LogError($"Profiler not enabled or memoryUsage was null, MemoryUsage not in usable state.");
+                    return;
+                }
+
+                memoryUsage.TotalAllocatedMemoryDelta += Profiler.GetTotalAllocatedMemoryLong() - startingAllocatedMemory;
+                memoryUsage.TotalReservedMemoryDelta += Profiler.GetTotalReservedMemoryLong() - startingReservedMemory;
+                memoryUsage.TotalUnusedReservedMemoryDelta += Profiler.GetTotalUnusedReservedMemoryLong() - startingUnusedMemory;
+            }
```
matthejo: The IDisposable pattern rules state that it should be safe to call Dispose multiple times without adverse side-effects. I think this should have a guard against that to prevent double-incrementing the memory allocation.  There's already a DisposableBase that correctly implements IDisposable in the SpatialAlignment part of SpectatorView, perhaps it should be moved to a place where this class can derive from it.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceMonitor.cs
```diff
@@ -7,17 +7,37 @@
...
+using UnityEngine.Profiling;
 
 namespace Microsoft.MixedReality.SpectatorView
 {
     public class StateSynchronizationPerformanceMonitor : Singleton<StateSynchronizationPerformanceMonitor>
     {
-        private struct StopWatchKey
+        public class MemoryUsage
+        {
+            public long TotalAllocatedMemoryDelta;
```
matthejo: >TotalAllocatedMemoryDelta [](start = 24, length = 25)

I'm not sure the "Delta" is necessary in this name. If I understand correctly, this represents the total allocated memory by some specific component/scope that's being measured. Is that correct?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetService.cs
```diff
@@ -16,10 +16,8 @@ internal class AssetService : Singleton<AssetService>, IAssetCache
 
         private readonly Dictionary<ShortID, IAssetSerializer<Texture>> textureSerializers = new Dictionary<ShortID, IAssetSerializer<Texture>>();
 
-        protected override void Awake()
+        protected void Start()
```
matthejo: >protected void Start() [](start = 8, length = 22)

I think this should be made virtual as well

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -224,27 +228,35 @@ internal abstract class AssetCache<TAssetEntry, TAsset> : AssetCache, IAssetCach
...
         public TAsset GetAsset(AssetId assetId)
         {
-            TAssetEntry assetEntry;
-            if (LookupByAssetId.TryGetValue(assetId, out assetEntry))
-            {
-                return assetEntry.Asset;
-            }
-            else
+            using (StateSynchronizationPerformanceMonitor.Instance.IncrementEventDuration(this.GetType().Name, "GetAsset"))
+            using (StateSynchronizationPerformanceMonitor.Instance.MeasureEventMemoryUsage(this.GetType().Name, "GetAsset"))
```
matthejo: >using (StateSynchronizationPerformanceMonitor.Instance.MeasureEventMemoryUsage(this.GetType().Name, "GetAsset")) [](start = 12, length = 112)

It would be surprising to me if this block or the GetAssetId block cause significant memory allocations. Is there some large allocation hidden here I'm not seeing?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/GlobalShaderPropertiesBroadcaster.cs
```diff
@@ -15,10 +15,8 @@ internal class GlobalShaderPropertiesBroadcaster : Singleton<GlobalShaderPropert
         private List<GlobalMaterialPropertyAsset> changedProperties = new List<GlobalMaterialPropertyAsset>();
         private CustomShaderPropertyAssetCache assetCache;
 
-        protected override void Awake()
+        protected void Start()
         {
```
matthejo: Add virtual to this as well

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceMonitor.cs
chrisfromwork: Ideally yes, but if there are subsequent allocations that occur at the same time you can end up with values that don't truly reflect the measure you were going for. I'm not sure if the delta in the name helps reflect this.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
chrisfromwork: I can remove these for now, I was using these events for comparison when defer loading assets.

chrisfromwork:
@stephenatwork ReadMessage isn't called on the user machine, its called by the editor/spectator machine upon receiving a message. The user decides to send this information on whether EnablePerformanceReporting is enabled.

We could start queuing this information, but some events may be called quite frequently. We'd have to ensure that at least some sort of cleanup step is ran every once in a while to free unneeded events. But IMO this is beyond the scope of this review.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
chrisfromwork: I think at this point in time i'll hold off on a combined method.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceMonitor.cs
chrisfromwork: This seems like a good suggestion, i will look into this now.

chrisfromwork: At this point in time, we could ditch the stopwatch allocations and use one static stopwatch, but there would like be some sort of class object that is handed to the TimeScope and MemoryScope every event call. I've opened the following issue for this:
https://github.com/microsoft/MixedReality-SpectatorView/issues/288

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationPerformanceMonitor.cs
chrisfromwork: DisposableBase is a class while this is a struct. I'll add a flag for now to check that the calculation isn't reperformed.

PR #287, opened 2019-11-26
Perf & button fixes for Build2019 demo scene

julenka:
# Overview
The Build2019 scene was running at 20fps when disconnected from SpectatorView because its camera settings were set to "ultra" quality, which enabled features like anti-aliasing that caused GPU to spend ~30 ms even when it was rendering no content. 

Fix this by using the default HoloLens2 camera profile in the Build2019 demo scene. Also fix several other ux issues I found in the scene such as the buttons not fully compressing.

# Changes
- Fixed: #285 by using the default HoloLens2 camera profile
- Update the button max depth and press distance so buttons press farther
- Hook up slider value update script so text field updates


src/SpectatorView.Unity/Assets/SpectatorView/Materials/BlueButton.png.meta
```diff
@@ -119,7 +119,7 @@ TextureImporter:
     outline: []
     physicsShape: []
     bones: []
-    spriteID: 
+    spriteID: 9b3ef1ad36c71cb43a93ce922c44f2a6
```
chrisfromwork: It looks like a lot of meta files have been updated here. Can they be reverted or is this changed needed?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
```diff
@@ -104,7 +104,7 @@ public override void Read(SocketEndpoint sendingEndpoint, BinaryReader message)
                 TextMeshObserver.fontSizeMax = message.ReadSingle();
                 TextMeshObserver.fontSizeMin = message.ReadSingle();
                 TextMeshObserver.fontStyle = (FontStyles)message.ReadInt32();
-                TextMeshObserver.fontWeight = message.ReadInt32();
+                TextMeshObserver.fontWeight = (FontWeight)message.ReadInt32();
```
chrisfromwork: What version of text mesh pro does the MRTK formally support? Text mesh pro versioning is independent of Unity versions and is managed through UPM. Technically, you could have fixed this issue by going to a lower version of text mesh pro in UPM.

chrisfromwork:
I'd prefer to get the meta files fixed before checking this in, but the review looks good to go after that.

src/SpectatorView.Unity/Assets/SpectatorView/Materials/BlueButton.png.meta
julenka: Those meta files get changed every time I open Unity, which I think means they are out of date. Somebody should fix the meta files. However, I reverted them to keep this PR focused.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
julenka: I believe MRTK supports the latest version, so I'd recommend keeping this in. Though technically MRTK does not mention which version they support, which is why I assumed latest.

julenka:
Thanks for the review, Chris! I reverted the meta files.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/NetworkedComponents/TextMeshPro/TextMeshProObserverBase.cs
chrisfromwork: Sounds good

PR #292, opened 2019-12-04
Remove unnecessary monobehaviour inheritance for AndroidRecordingService and iOSRecordingService

chrisfromwork:
This review updates our recording services to no longer be Monobehaviours. They didn't use any MonoBehaviour functionality and the fact that they were MonoBehaviour was generating warnings at runtime when SpectatorView created them.

## Breaking Change Details

###Notes:
We are changing AndroidRecordingService and iOSRecordingService to no longer be monobehaviours based on some warnings that occurred during initialization. This may cause issues for anyone inheriting from these classes.

### Migration Instructions:
1. Convert any classes inheriting from AndroidRecordingService or iOSRecordingService to no longer be monobehaviours.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/ScreenRecording/AndroidRecordingService.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/ScreenRecording/AndroidRecordingService.cs
@@ -9,8 +9,7 @@ namespace Microsoft.MixedReality.SpectatorView
     /// <summary>
     /// Class implementing <see cref="IRecordingService"/> for the Android platform
     /// </summary>
-    public class AndroidRecordingService : MonoBehaviour,
-    IRecordingService
+    public class AndroidRecordingService : IRecordingService
     {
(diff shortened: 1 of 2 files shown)


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/ScreenRecording/AndroidRecordingService.cs
```diff
@@ -9,8 +9,7 @@ namespace Microsoft.MixedReality.SpectatorView
     /// <summary>
     /// Class implementing <see cref="IRecordingService"/> for the Android platform
     /// </summary>
-    public class AndroidRecordingService : MonoBehaviour,
-    IRecordingService
+    public class AndroidRecordingService : IRecordingService
```
wiwei: >AndroidRecordingService [](start = 17, length = 23)

If you think that this is something that has a chance of being extended, adding a release note for this would be helpful.

Likelihood may be low so it's probably fine to also not have anything mentioned

Issue #197, opened 2019-08-22
Spectator view should defer loading large assets until needed.


chrisfromwork:
Spectator view loads asset caches initially, so if a lot of meshes/large memory assets are included in asset caches, you can hit a point where the app crashes based on a lack of memory. We should assess whether loading these assets can be deferred until they are needed.

emartinDesignInteractive:
Any update on this? It is preventing me from progressing using Spectator View

chrisfromwork:
We are looking into this now.

RealFrogee:
After I run "Update All Asset Caches" in my project, the project build size has increased by 1G. And when I open it on Hololens,it crashed.I have tested several situations.Without "Update All Asset Caches" the build can run normally on Hololens.
Is the crash caused by the large assets problem?Is there a solution now?

chrisfromwork:
@RealFrogee the issue is based on how asset caches currently work. Basically all assets in your unity project get declared in an asset cache. Then when the hololens starts up it tries to load all asset caches. This is okay for small unity projects/targetted demos, but is really bad for real world applications whose projects contain a lot of content. The review we have out in #295 may be able to unblock you. It still has the same amount of asset references in a resource folder, which i would imagine will still increase your project build size. But compared to attempting to load all assets on start, a list of asset names is loaded. Then when spectator view observes assets declared in your scene, this asset name list is used to load individual asset references. There will still be some caveats, if you have any assets with the same name and type, all of them will be loaded at the same time even if you are only using one of them.

PR #295, opened 2019-12-04
Defer loading assets in asset caches until needed

chrisfromwork:
This change addresses #197

All assets in a Unity project get declared in AssetCaches and loaded regardless of whether or not they are used by Spectator View's state synchronization components. This can result in applications running out of memory if the Unity project was quite large. This change defers loading assets until they are required. The pull request includes the following:

1. AssetCaches now contain entries comprised of asset names and asset ids and are loaded on start.
2. Asset references are now declared in new files created for each asset name.
3. When a request is made to the asset cache for an asset or asset id, a lookup process occurs using an asset name/asset id table. If the asset isn't found, it is then loaded from one of the new files based on the asset name.
4. Stopwatch performance events now create samples that can be viewed in the unity profiler
5. A new flag is added that causes the application to load all assets during initialization for scenarios where deferred asset loading isn't preferred.
6. Previous asset caches are deleted from the repository.

## Regressions/Changes in Behavior:
We will now use Asset names to look up Assets in Asset Caches. Some Assets, such as Textures, can be renamed at runtime, which can break Asset lookup. We do not think this is a common developer scenario, but it is an identified change in behavior.

### Notes:
This deferred loading setup changes the content of asset caches, Asset references are now relocated to a new location/new files. For the new logic to work, you will need to update the asset caches in your application.

### Migration Instructions:
1. Delete any previously existing asset caches. These will be found in the Generated.StateSynchronization.AssetCaches folder
2. Create your asset caches by running `Spectator View -> Update All Asset Caches`


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -277,35 +394,117 @@ public override void ClearAssetCache()
...
+        }
+
+        string GetValidAssetName(string name)
+        {
+            if (name == null)
+            {
+                return null;
+            }
+
+            return name.Replace(":", "_");
```
matthejo: >return name.Replace(":", "_"); [](start = 12, length = 30)

This needs to replace any invalid path character in a name with some other valid character, including "/", "\", etc. You can use Path.GetInvalidPathChars to get the complete list

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -277,35 +394,117 @@ public override void ClearAssetCache()
...
+                var entry = new AssetCacheEntry();
+                entry.Asset = asset;
+                entry.AssetId = assetId;
+                lookupByAsset[asset] = entry;
+                lookupByAssetId[assetId] = entry;
             }
 
-            CleanUpUnused(oldAssets, unvisitedAssets);
-            assets = oldAssets.Values.OrderBy(x => x.AssetId.FileIdentifier).ThenBy(x => x.AssetId.Guid).ToArray();
+            foreach(var idPair in unvisitedIds)
```
matthejo: >foreach [](start = 12, length = 7)

Nit: spacing between "foreach" and "(" here and below

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -277,35 +394,117 @@ public override void ClearAssetCache()
...
+                        Debug.LogError($"AssetId did not have a registered Asset: {assetId}");
+                        continue;
+                    }
+
+                    assetCacheEntries.Add(entry);
+                }
+
+                content.AssetCacheEntries = assetCacheEntries.ToArray();
+                Debug.Log($"Creating asset: {content} {assetName}");
+                AssetDatabase.CreateAsset(content, assetName);
```
matthejo: It seems like something needs to delete old assets from the asset cache directory. Otherwise, every time an asset name changes, the old version of the asset will still be in this directory

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/HoloLensTrackingObserver.cs
```diff
@@ -15,15 +15,15 @@ public override TrackingState TrackingState
         {
             get
             {
-#if UNITY_WSA
+#if UNITY_EDITOR
```
wiwei: >UNITY_EDITOR [](start = 4, length = 12)

Was this mainly a change to get alphabetical ordering for the ifdefs? Functionally it seems the same (AFAICT)

wiwei:
Meta question on this thing, which is that it's good we have a breaking change/migration note in this PR - are we tracking all of the changes in some other issue (i.e. so next release we don't need to do a scrape, and can just copy from that issue?).

On the MRTK side of the world it's been helpful to update the changelog as we do it, so that we don't have to scrape at the last moment

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/HoloLensTrackingObserver.cs
chrisfromwork: it feels like this may be a product of a weird merge. I do not know why this changed.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/README.md
```diff
@@ -48,6 +48,19 @@ Event durations will vary based on the content you have defined in your scene. W
...
 2) When you leave all of the options as `Update Continuously` in the global performance settings, you will see material property updates reported in the Event Count section of the performance window. Viewing material updates here may better inform what material properties to declare in the `Material Property Overrides` within your global performance settings when changing `Material Properties` to only `Update Once on Start`.
 
+### Profiling Performance Bottlenecks
+Spectator View state synchronization component events can be viewed in the Unity Profiler. This can help generate a better understanding of the frame by frame impact associated with state synchronization. To have Spectator View events appear in the profiler, do the following:
+
+1) Open `Spectator View -> Edit Settings` and check `Force Performance Reporting`.
+
+![Marker](../../../../../../doc/images/ForcePerformanceReporting.png)
+
+2) Create a `Development Build` with `Autconnect Profiler` enabled. This will allow the Unity Profiler window to attach to your device.
```
wiwei:
```suggestion
2) Create a `Development Build` with `Autoconnect Profiler` enabled. This will allow the Unity Profiler window to attach to your device.
```

wiwei: I think that there are a few specific steps that need to be taken to get the profiler network connection to work, which are described here:

https://docs.unity3d.com/2018.2/Documentation/Manual/windowsstore-profiler.html

It's possible that we've omitted them here because the act of hitting "Hololens" on the platform switcher actually sets this for the user, so maybe it's okay to not mention it? Maybe the ask here is to have a little sentence like...

Create a `Development Build` with `Autconnect Profiler` enabled. This will allow the Unity Profiler window to attach to your device. (For more information on Unity Profiler, including prerequisite setup steps, check out the [Unity documentation](https://docs.unity3d.com/2018.2/Documentation/Manual/windowsstore-profiler.html)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -8,29 +8,63 @@
...
         }
 
-        public static string GetAssetPath(string assetName, string assetExtension)
+        public static string GetAssetCachePath(string assetName, string assetExtension)
         {
-            return $"Assets/{assetCacheDirectory}/Resources/" + assetName + assetExtension;
+            return $"Assets/{assetCacheDirectory}/Resources/{assetName}{assetExtension}";
+        }
+
+        public static string GetAssetCachesContentPath(string assetName, string assetExtension)
```
wiwei: Nit: Can you add some comments about what these functions do (i.e. especially the two of them together being so close, taking the same set of parameters, it's not clear as a reader when I would call one and when I'd call the other).

Especially because this is the base class for a lot of other stuff, I think it's important to have these sorts of things documented for posterity's sake.

I think what I'm really looking for is documentation about what a "content" path is and a "regular path" is.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -178,83 +220,158 @@ public static IEnumerable<T> EnumerateAllComponentsInScenesAndPrefabs<T>()
...
     }
 
-    internal abstract class AssetCache<TAssetEntry, TAsset> : AssetCache, IAssetCache
-        where TAssetEntry : AssetCacheEntry<TAsset>, new()
+    internal abstract class AssetCache<TAsset> : AssetCache, IAssetCache
         where TAsset : UnityEngine.Object
     {
-        [SerializeField]
-        private TAssetEntry[] assets = null;
+        const string assetContentExtension = ".asset";
```
wiwei: >assetContentExtension [](start = 21, length = 21)

I think we should stick to Pascal case: 

[https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions)
[https://blogs.msdn.microsoft.com/mschray/2013/05/06/practical-naming-standards-for-c/](https://blogs.msdn.microsoft.com/mschray/2013/05/06/practical-naming-standards-for-c/)

So AssetContentExtensions.

Especially for constants

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -178,83 +220,158 @@ public static IEnumerable<T> EnumerateAllComponentsInScenesAndPrefabs<T>()
...
 
-        private Dictionary<AssetId, TAssetEntry> lookupByAssetId;
-        private Dictionary<TAsset, TAssetEntry> lookupByAsset;
+#pragma warning disable 414
```
wiwei: >414 [](start = 24, length = 3)

Non-nit: every time we disable a warning, we should have a comment about why the warning is being ignored/disabled (i.e. why it's safe to ignore)

wiwei: We have to set a good example for the future where people may copy-paste things willy-nilly, and commentless-warning disables can proliferate inadvertently this way.

---
In reply to: [355037489](https://github.com/microsoft/MixedReality-SpectatorView/pull/295#discussion_r355037489) [](ancestors = 355037489)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -178,83 +220,158 @@ public static IEnumerable<T> EnumerateAllComponentsInScenesAndPrefabs<T>()
...
+        [SerializeField]
+        private NameEntry[] assets = null;
+#pragma warning restore 414
 
-        protected IDictionary<AssetId, TAssetEntry> LookupByAssetId
+        private Dictionary<string, List<AssetId>> LookupByName
```
wiwei: >LookupByName [](start = 50, length = 12)

Just to check (haven't read through all the code yet, so I'm leaving this comment with the knowledge it might get resolved after reading,  lookupByName gets populated later down the line correctly as "assets" get filled right? i.e. there's no danger of accessing LookupByName early (before assets is non-null?)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -178,83 +220,158 @@ public static IEnumerable<T> EnumerateAllComponentsInScenesAndPrefabs<T>()
...
+#pragma warning restore 414
+
         public TAsset GetAsset(AssetId assetId)
         {
             using (StateSynchronizationPerformanceMonitor.Instance.IncrementEventDuration(this.GetType().Name, "GetAsset"))
             {
-                TAssetEntry assetEntry;
-                if (LookupByAssetId.TryGetValue(assetId, out assetEntry))
+                if (assetId == null ||
+                assetId == AssetId.Empty ||
```
wiwei: >assetId == AssetId.Empty || [](start = 16, length = 27)

Nit: I think there are some missing indents here (i.e. bring assetId on the next level to at least where assetId is on the preceding one (and same for the next line).

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -178,83 +220,158 @@ public static IEnumerable<T> EnumerateAllComponentsInScenesAndPrefabs<T>()
...
+                assetId.Name == string.Empty)
+                {
+                    return default(TAsset);
```
wiwei: >return [](start = 20, length = 6)

Curious - is it interesting to increment the event duration if the passed in AssetId is invalid? Or is it separately interesting to log that?

Otherwise I assume you can end up getting some zero-duration (or near-zero duration) events which are probably interesting to know separately (i.e. something is "wrong" because an invalid parameter was passed, rather than just because the duration was effectively zero)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -178,83 +220,158 @@ public static IEnumerable<T> EnumerateAllComponentsInScenesAndPrefabs<T>()
...
+                }
+
+                if (lookupByAssetId.TryGetValue(assetId, out var entry))
```
wiwei: For the performance monitor (again, I don't have a lot of background here, so maybe this question isn't fully valid), is it interesting to bucketize based on cache misses and cache hits? i.e. is the event duration in total interesting, or are we really interested in the real load time?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -178,83 +220,158 @@ public static IEnumerable<T> EnumerateAllComponentsInScenesAndPrefabs<T>()
...
 
         public AssetId GetAssetId(TAsset asset)
         {
             using (StateSynchronizationPerformanceMonitor.Instance.IncrementEventDuration(this.GetType().Name, "GetAssetId"))
             {
-                TAssetEntry assetEntry;
-                if (asset != null && LookupByAsset.TryGetValue(asset, out assetEntry))
+                if (asset == null ||
+                    asset.name == null ||
+                    asset.name == string.Empty)
```
wiwei: For "simplicity" (or maybe I dunno what the right word is), consider:

[https://docs.microsoft.com/en-us/dotnet/api/system.string.isnullorempty?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.string.isnullorempty?view=netframework-4.8)

I think same for the stuff up above too

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -178,83 +220,158 @@ public static IEnumerable<T> EnumerateAllComponentsInScenesAndPrefabs<T>()
...
+        public override void LoadAllAssets()
+        {
+            using (StateSynchronizationPerformanceMonitor.Instance.IncrementEventDuration(this.GetType().Name, "LoadAllAssets"))
+            {
+                bool successful = true;
+                foreach (var pair in LookupByName)
+                {
+                    if (!TryLoadAssets(pair.Key))
+                    {
+                        successful = false;
```
wiwei: > successful = false; [](start = 23, length = 20)

Consider logging what the failed asset was (i.e. so instead of seeing "something failed" you'll see "this failed")

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -178,83 +220,158 @@ public static IEnumerable<T> EnumerateAllComponentsInScenesAndPrefabs<T>()
...
+                    Debug.LogWarning($"Not all assets were found for {this.GetType().Name} during forced asset loading");
+                }
+            }
+        }
+
+        private bool TryLoadAssets(string name)
+        {
+            using (StateSynchronizationPerformanceMonitor.Instance.MeasureEventMemoryUsage(this.GetType().Name, "LoadingAssets"))
+            {
+                if (!LookupByName.TryGetValue(name, out var assetsIds))
```
wiwei: >if (!LookupByName.TryGetValue(name, out var assetsIds)) [](start = 16, length = 55)

From a calling expectation perspective - if this has failed, is this actually an expected case, or does this reflect a failure on the caller (or something else earlier) (i.e. is it worth logging that an invalidly named asset/not found asset was attempted to be found?)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -178,83 +220,158 @@ public static IEnumerable<T> EnumerateAllComponentsInScenesAndPrefabs<T>()
...
+                {
+                    return false;
+                }
+
+                string assetPath = $"{assetsFolderName}/{GetValidAssetName(name)}_{this.GetType().Name}";
```
wiwei: >GetValidAssetName(name)}_{this.GetType().Name} [](start = 57, length = 46)

Consider making this into a helper function, given that this specific structure is used on line 350 and on line 387)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -277,35 +404,129 @@ public override void ClearAssetCache()
...
+            HashSet<Tuple<string, AssetId>> unvisitedIds = new HashSet<Tuple<string, AssetId>>();
+            lookupByAsset = new Dictionary<TAsset, AssetCacheEntry>();
+            lookupByAssetId = new Dictionary<AssetId, AssetCacheEntry>();
+            foreach (var listPair in oldAssets)
+            {
+                foreach (var id in listPair.Value)
+                {
+                    unvisitedIds.Add(new Tuple<string, AssetId>(listPair.Key, id));
+                }
+            }
```
wiwei: This is a medium-sized function that has non-trivial data-reconciliation steps - usually when I see these sorts of things, it's helpful if there's some comments/overall section that explains the overall algorithm that's running through to set the expectations for what the code is doing.

The ask here isn't document line-by-line what is happening (i.e. it's obvious that we're visiting old assets and adding them to a hashset for reconciliation), but more an overall set of steps.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -277,35 +404,129 @@ public override void ClearAssetCache()
...
 
             foreach (TAsset asset in EnumerateAllAssets())
             {
-                unvisitedAssets.Remove(asset);
-                if (!oldAssets.ContainsKey(asset))
+                if (asset.name == null ||
+                    asset.name == string.Empty ||
+                    !AssetDatabase.TryGetGUIDAndLocalFileIdentifier(asset, out string guid, out long localid))
```
wiwei: >localid [](start = 101, length = 7)

Super Nit: localId

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -277,35 +404,129 @@ public override void ClearAssetCache()
...
+                    if (ids.Count == 0)
                     {
-                        oldAssets.Add(asset, new TAssetEntry
-                        {
-                            AssetId = new AssetId(new Guid(guid), localid),
-                            Asset = asset
-                        });
+                        string assetName = GetAssetCachesContentPath($"{GetValidAssetName(idPair.Item1)}_{this.GetType().Name}", assetContentExtension);
+                        AssetDatabase.DeleteAsset(assetName);
                     }
```
wiwei: >} [](start = 20, length = 1)

If the count of the ID list is now zero, should we also simply remove the entry from oldAssets entirely? i.e. any reason to store an empty list?

wiwei:
>         /// Assets are ordered in their associated AssetCache by AssetId (first by file identifier, then by guid).

Is there also something to mention here about how they are sorted by name first?

---
Refers to: src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs:402 in 0a04138. [](commit_id = 0a0413826cd8909db0730aaa2b4d5c6a51f8afd0, deletion_comment = False)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -277,35 +404,129 @@ public override void ClearAssetCache()
...
+            if (name == null)
+            {
+                return null;
             }
 
-            CleanUpUnused(oldAssets, unvisitedAssets);
-            assets = oldAssets.Values.OrderBy(x => x.AssetId.FileIdentifier).ThenBy(x => x.AssetId.Guid).ToArray();
+            var invalidChars = Path.GetInvalidPathChars();
+            string updatedName = name;
+            foreach(var invalid in invalidChars)
```
wiwei: >foreach [](start = 12, length = 7)

Nit: foreach ( (instead of foreach(

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
```diff
@@ -277,35 +404,129 @@ public override void ClearAssetCache()
...
                 }
+
+                content.AssetCacheEntries = assetCacheEntries.ToArray();
+                Debug.Log($"Creating asset: {content} {assetName}");
+                AssetDatabase.CreateAsset(content, assetName);
+            }
+#endif
+        }
+
+        string GetValidAssetName(string name)
```
wiwei: >GetValidAssetName [](start = 15, length = 17)

Is it possible for this function to create naming collisions? i.e. because we're removing certain illegal characters (which may have made them unique before), they now become non-unique? If this happens, how will this manifest?

wiwei: >GetValidAssetName [](start = 15, length = 17)

Nit: Add comments for this function (i.e. returns name stripped of invalid Path characters (i.e. >, <... other stuff)

wiwei: Also, could maybe be static right

---
In reply to: [355049249](https://github.com/microsoft/MixedReality-SpectatorView/pull/295#discussion_r355049249) [](ancestors = 355049249)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/BroadcasterSettings.cs
```diff
@@ -37,5 +37,17 @@ public bool ForcePerformanceReporting
         {
             get { return forcePerformanceReporting; }
         }
+
+        [SerializeField]
+        [Tooltip("Check to force loading all assets in asset caches on initialization. Note: this will improve the performance of identifying assets for sending asset changes between devices, but it will require using more memory.")]
+        private bool forceLoadAllAssetsDuringInitialization = false;
+
+        /// <summary>
+        /// Check to force loading all assets in asset caches on initialization. Note: this will improve the performance of identifying assets for sending asset changes between devices, but it will require using more memory.
```
wiwei: >Check to force [](start = 12, length = 14)

Nit: check to force -> forces

This is mirror the language of forcePerformanceReporting, and also because for getters there's nothing to check.

chrisfromwork:
we don't have something like this currently. I have been flagging everything as breaking change for scraping.

---
In reply to: [562739082](https://github.com/microsoft/MixedReality-SpectatorView/pull/295#issuecomment-562739082) [](ancestors = 562739082)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/README.md
chrisfromwork: This is a good suggestion.

---
In reply to: [355034044](https://github.com/microsoft/MixedReality-SpectatorView/pull/295#discussion_r355034044) [](ancestors = 355034044)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
chrisfromwork: sounds good.

---
In reply to: [355037299](https://github.com/microsoft/MixedReality-SpectatorView/pull/295#discussion_r355037299) [](ancestors = 355037299)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
chrisfromwork: We can already end up with non-unique name collisions. We currently load all assets with the same name that are of the same type at the same time. Once loaded, they will have different asset guids/file identifiers. 

---
In reply to: [355049066](https://github.com/microsoft/MixedReality-SpectatorView/pull/295#discussion_r355049066) [](ancestors = 355049066)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
chrisfromwork: good catch

---
In reply to: [355047556](https://github.com/microsoft/MixedReality-SpectatorView/pull/295#discussion_r355047556) [](ancestors = 355047556)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
chrisfromwork: This likely warrants an error.

---
In reply to: [355042828](https://github.com/microsoft/MixedReality-SpectatorView/pull/295#discussion_r355042828) [](ancestors = 355042828)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
chrisfromwork: TryLoadAssets already flags what portion of the look up failed. I don't think we need to duplicate that information here, but we could still keep this asset lookup failure for the overall asset cache.

---
In reply to: [355041870](https://github.com/microsoft/MixedReality-SpectatorView/pull/295#discussion_r355041870) [](ancestors = 355041870)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
chrisfromwork: for now that measure is primarily to understand how much time in total is spent in this function. We could have different event logging for found vs not found, but I don't think that is needed for now.

---
In reply to: [355039776](https://github.com/microsoft/MixedReality-SpectatorView/pull/295#discussion_r355039776) [](ancestors = 355039776)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
chrisfromwork: we were just using the event duration in total to understand whether or not this new deferred asset loading strategy had a large impact on the app responsiveness compared to the old asset loading method. I think as more time passes we'll have to make this performance reporting better, but for now this event has served its purpose.

---
In reply to: [355040093](https://github.com/microsoft/MixedReality-SpectatorView/pull/295#discussion_r355040093) [](ancestors = 355040093)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/AssetCache.cs
chrisfromwork: I'll add some error logging here. assets being null can't be fixed at runtime. These AssetCaches are stored in the Resources folder so assets should already be populated when this AssetCache is created.

---
In reply to: [355038927](https://github.com/microsoft/MixedReality-SpectatorView/pull/295#discussion_r355038927) [](ancestors = 355038927)

PR #305, opened 2019-12-18
Update INetworkConnection's send definition based on review comments.

chrisfromwork:
We change the send definition to:

`void Send(byte[] data, long offset, long length);`

### DIFF
--- a/src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CameraPoseProvider.cs
+++ b/src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CameraPoseProvider.cs
@@ -117,7 +117,7 @@ namespace Microsoft.MixedReality.SpectatorView
                 message.Write(cameraRotation);
 
                 var data = stream.ToArray();
-                networkManager.Broadcast(ref data);
+                networkManager.Broadcast(stream.GetBuffer(), 0, stream.Position);
             }
         }
 
--- a/src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CameraPoseProvider.cs
+++ b/src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CameraPoseProvider.cs
@@ -115,8 +115,8 @@ namespace Microsoft.MixedReality.SpectatorView
                 message.Write(timestamp);
                 message.Write(cameraPosition);
                 message.Write(cameraRotation);
+                message.Flush();
 
-                var data = stream.ToArray();
                 networkManager.Broadcast(stream.GetBuffer(), 0, stream.Position);
             }
         }
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Networking/INetworkConnection.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Networking/INetworkConnection.cs
@@ -30,8 +30,10 @@ namespace Microsoft.MixedReality.SpectatorView
         /// <summary>
         /// Call to send data to the connected peer
         /// </summary>
-        /// <param name="data">a reference to the data to send, data will be set to null after a send/the INetworkConnection will take over ownership of the array</param>
-        void Send(ref byte[] data);
+        /// <param name="data">A reference to the data to send</param>
+        /// <param name="offset">The offset from the start of the array to use to obtain the data to send</param>
+        /// <param name="length">The length of the data to send</param>
+        void Send(byte[] data, long offset, long length);
 
         /// <summary>
         /// Returns a string that can be used to identify the network connection in UI/for logging.
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Socketer/TCPNetworkConnection.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Socketer/TCPNetworkConnection.cs
@@ -76,7 +76,7 @@ namespace Microsoft.MixedReality.SpectatorView
         }
 
         /// <inheritdoc />
-        public void Send(ref byte[] data)
+        public void Send(byte[] data, long offset, long length)
         {
             if (!IsConnected)
             {
@@ -86,15 +86,15 @@ namespace Microsoft.MixedReality.SpectatorView
             {
                 try
                 {
-                    socketerClient.SendNetworkMessage(data, sourceId);
+                    var payload = new byte[length];
+                    Array.Copy(data, offset, payload, 0, length);
+                    socketerClient.SendNetworkMessage(payload, sourceId);
                 }
                 catch
                 {
                     socketerClient.Disconnect(sourceId);
                 }
             }
-
-            data = null;
         }
 
         /// <inheritdoc />
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemParticipant.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemParticipant.cs
@@ -169,6 +169,7 @@ namespace Microsoft.MixedReality.SpectatorView
                 if (previousCoordinateStatusMessage == null || !previousCoordinateStatusMessage.SequenceEqual(newCoordinateStatusMessage))
                 {
                     NetworkConnection.Send(newCoordinateStatusMessage, 0, newCoordinateStatusMessage.Length);
+                    previousCoordinateStatusMessage = newCoordinateStatusMessage;
                 }
             }
         }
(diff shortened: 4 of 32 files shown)


src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CameraPoseProvider.cs
```diff
@@ -117,7 +117,7 @@ private void SendCameraPose(float timestamp, Vector3 cameraPosition, Quaternion
                 message.Write(cameraRotation);
 
                 var data = stream.ToArray();
```
ivte-ms: Unused?

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CameraPoseProvider.cs
```diff
@@ -117,7 +117,7 @@ private void SendCameraPose(float timestamp, Vector3 cameraPosition, Quaternion
...
-                networkManager.Broadcast(ref data);
+                networkManager.Broadcast(stream.GetBuffer(), 0, stream.Position);
```
ivte-ms: In some other places you flush the writer. Is this a required step here as well, or will it happen automatically when you call GetBuffer() (in which case other flushes can probably be removed)?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationBroadcaster.cs
```diff
@@ -181,8 +180,7 @@ public void OnFrameCompleted()
                     message.Write(StateSynchronizationObserver.PerfCommand);
                     StateSynchronizationPerformanceMonitor.Instance.WriteMessage(message, numFrames);
                     message.Flush();
-                    var data = memoryStream.ToArray();
-                    connectionManager.Broadcast(ref data);
+                    connectionManager.Broadcast(memoryStream.GetBuffer(), 0, memoryStream.Position);
```
ivte-ms: Adding a helper method that accepts a memory stream could be a better idea than repeating the same pattern multiple times (especially if Flush is actually required, see the other comments).

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemParticipant.cs
```diff
@@ -163,12 +163,12 @@ private void SendCoordinateStateMessage()
 
                 message.Write(position);
                 message.Write(rotation);
+                message.Flush();
 
                 byte[] newCoordinateStatusMessage = stream.ToArray();
```
ivte-ms: You probably don't need to allocate an array to perform the checks below.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/DeviceInfoBroadcaster.cs
```diff
@@ -59,8 +59,7 @@ private void SendDeviceInfo()
                 message.Write(GetMachineName());
                 message.Write(GetIPAddress());
 
-                var data = memoryStream.ToArray();
-                networkManager.Broadcast(ref data);
+                networkManager.Broadcast(memoryStream.GetBuffer(), 0, memoryStream.Position);
```
ivte-ms: Another missing flush?

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CameraPoseProvider.cs
chrisfromwork: I'm not sure what ends up happening. In my experience flush's haven't been needed for the payloads weve sent around here.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemParticipant.cs
chrisfromwork: the allocation seems more important for caching the information which it looks like we were failing to do here.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationBroadcaster.cs
chrisfromwork: I'll hold off on this for now. I've opened the following issue to look into this: https://github.com/microsoft/MixedReality-SpectatorView/issues/307

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CameraPoseProvider.cs
chrisfromwork: The documentation here is a bit vague. I'm not sure anyone would ever extend BinaryWriter here https://docs.microsoft.com/en-us/dotnet/api/system.io.binarywriter.flush?view=netframework-4.8 We also don't use any UTF7 characters right now. That doesn't necessarily mean that someone wouldn't in the future.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/StateSynchronization/StateSynchronizationBroadcaster.cs
AppleBeam: I don't think #307 is related. It's not about reusing the streams or writers, it's about making sure that the same error-prone pattern is not repeated multiple times. I don't think it's a blocker if all the usages are fixed here, though.

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CameraPoseProvider.cs
ivte-ms: Would be nice to make some decision here (either delete all unnecessary flushes or make sure that it's hard to forget it).

stephenatwork: @chrisfromwork 
Flush just calls Stream.Flush()
MemoryStream.Flush() is a no-op.
So it seems clear to delete the flushes.

https://referencesource.microsoft.com/#mscorlib/system/io/binarywriter.cs
https://referencesource.microsoft.com/#mscorlib/system/io/memorystream.cs

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Socketer/TCPNetworkConnection.cs
```diff
@@ -86,15 +86,15 @@ public void Send(ref byte[] data)
             {
                 try
                 {
-                    socketerClient.SendNetworkMessage(data, sourceId);
+                    var payload = new byte[length];
+                    Array.Copy(data, offset, payload, 0, length);
+                    socketerClient.SendNetworkMessage(payload, sourceId);
```
stephenatwork: Beyond the scope of this review, but are we sure all the small writes are getting coalesced and we're not spamming lots of packets? A quick look in a packet sniffer could be useful.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Networking/INetworkConnection.cs
```diff
@@ -30,8 +30,10 @@ public interface INetworkConnection
         /// <summary>
         /// Call to send data to the connected peer
         /// </summary>
-        /// <param name="data">a reference to the data to send, data will be set to null after a send/the INetworkConnection will take over ownership of the array</param>
-        void Send(ref byte[] data);
+        /// <param name="data">A reference to the data to send</param>
+        /// <param name="offset">The offset from the start of the array to use to obtain the data to send</param>
+        /// <param name="length">The length of the data to send</param>
+        void Send(byte[] data, long offset, long length);
```
fibann: As far as I understood, Span seems the way to go for memory spans and is available in Unity with a NuGet package. So why not Span? I am sure this has been discussed but it would be good to have the decision and rationale in writing.

stephenatwork: The issue is that support for consuming nuget packages in unity won't arrive until after the networking abstraction ships.
Options discussed included copy/pasting part of the nuget package. But the simplest is to use this interface and accept a minor breaking change when span becomes available. (byte[],int,int) will be fairly painless to migrate to span, (ref byte[]) less so.

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CameraPoseProvider.cs
chrisfromwork: checking this in for now, we can remove these moving forward as we redefine this interface again for spans

PR #306, opened 2019-12-18
Improve the building experience for the Spectator View Compositor Native Plugin

chrisfromwork:
Most users likely won't have both an elgato capture card and a blackmagic capture card. However, our current codebase requires obtaining dependencies to build for both solutions. Without a blackmagic capture card you won't have a device serial number required for getting the developer sdk. Therefore, we should make our native project friendlier to users that only have one capture card. This review contains the following changes:

1) Blackmagic and elgato frame provider code is now hidden behind INCLUDE_BLACKMAGIC and INCLUDE_ELGATO preprocessor definitions.
2) dependencies.props now enables the INCLUDE_BLACKMAGIC and INCLUDE_ELGATO preprocessor definitions based on whether the user has provided valid paths for the capture card.
3) If no capture card dependency has been provided, an error is generated during the prebuild setup based on a target added to the dependencies.props files.
4) Our scripts for building native plugins locally have been updated to not require blackmagic sdks.
5) Submodule dependencies have been added for the elgato and vcpkg projects. Both projects are mit license.
6) The compositor now exposes to the unity editor what capture cards are supported by the compiled dll. This should help decrease confusion if the dll is miscompiled for the users desired setup.

### Notes:
This is a minimal breaking change. Unless someone had written their own compositor ui interface, noone should need to update their own code. However, this change exposes a new function that the unity editor compositor needs, so a recompile of the compositor dll is needed.

### Migration Instructions:
1. Recompile SpectatorView.Compositor.dll and SpectatorView.Compositor.UnityPlugin.dll and readd them to your Unity project.

### DIFF
--- a/src/SpectatorView.Native/SpectatorView.Compositor/UnityPlugin/UnityCompositorInterface.cpp
+++ b/src/SpectatorView.Native/SpectatorView.Compositor/UnityPlugin/UnityCompositorInterface.cpp
@@ -343,6 +343,16 @@ UNITYDLL int GetVideoRecordingFrameHeight(VideoRecordingFrameLayout frameLayout)
     }
 }
 
+UNITYDLL bool IsFrameProviderSupported(int providerId)
+{
+	if (ci == nullptr)
+	{
+		ci = new CompositorInterface();
+	}
+
+	return ci->IsFrameProviderSupported((IFrameProvider::ProviderType) providerId);
+}
+
 UNITYDLL bool InitializeFrameProviderOnDevice(int providerId)
 {
     if (g_outputTexture == nullptr ||
--- a/src/SpectatorView.Native/SpectatorView.Compositor/UnityPlugin/UnityCompositorInterface.cpp
+++ b/src/SpectatorView.Native/SpectatorView.Compositor/UnityPlugin/UnityCompositorInterface.cpp
@@ -138,6 +138,12 @@ void UNITY_INTERFACE_API UnityPluginUnload()
     OnGraphicsDeviceEvent(kUnityGfxDeviceEventShutdown);
 
     DeleteCriticalSection(&lock);
+
+	if (ci != nullptr)
+	{
+		delete ci;
+		ci = nullptr;
+	}
 }
 
 UNITYDLL void UpdateCompositor()
--- a/src/SpectatorView.Native/SpectatorView.Compositor/dependencies.props
+++ b/src/SpectatorView.Native/SpectatorView.Compositor/dependencies.props
@@ -7,11 +7,22 @@
     <!-- From: https://www.blackmagicdesign.com/support -->
     <DeckLink_inc>C:\Blackmagic DeckLink SDK 10.9.11\Win\include</DeckLink_inc>
     <!-- From: https://github.com/elgatosf/gamecapture -->
-    <Elgato_Filter Condition="$(Elgato_Filter) == ''">C:\gamecapture\VideoCaptureFilter</Elgato_Filter>
+    <Elgato_Filter Condition="$(Elgato_Filter) == ''">C:\elgatosf\gamecapture\VideoCaptureFilter</Elgato_Filter>
   </PropertyGroup>
-  <PropertyGroup />
-  <ItemDefinitionGroup />
-  <ItemGroup />
+  <ItemDefinitionGroup>
+    <ClCompile Condition="Exists('$(DeckLink_inc)') And Exists('$(Elgato_Filter)')">
+      <PreprocessorDefinitions>INCLUDE_BLACKMAGIC;INCLUDE_ELGATO;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+    <ClCompile Condition="Exists('$(DeckLink_inc)') And !Exists('$(Elgato_Filter)')">
+      <PreprocessorDefinitions>INCLUDE_BLACKMAGIC;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+    <ClCompile Condition="!Exists('$(DeckLink_inc)') And Exists('$(Elgato_Filter)')">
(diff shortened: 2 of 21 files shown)


src/SpectatorView.Native/SpectatorView.Compositor/UnityPlugin/UnityCompositorInterface.cpp
```diff
@@ -343,6 +343,16 @@ UNITYDLL int GetVideoRecordingFrameHeight(VideoRecordingFrameLayout frameLayout)
     }
 }
 
+UNITYDLL bool IsFrameProviderSupported(int providerId)
+{
+	if (ci == nullptr)
+	{
+		ci = new CompositorInterface();
```
andreiborodin: How do we handle clean up of this heap allocation?

src/SpectatorView.Native/SpectatorView.Compositor/dependencies.props
```diff
@@ -7,11 +7,22 @@
...
+      <PreprocessorDefinitions>INCLUDE_BLACKMAGIC;INCLUDE_ELGATO;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+    <ClCompile Condition="Exists('$(DeckLink_inc)') And !Exists('$(Elgato_Filter)')">
+      <PreprocessorDefinitions>INCLUDE_BLACKMAGIC;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+    <ClCompile Condition="!Exists('$(DeckLink_inc)') And Exists('$(Elgato_Filter)')">
+      <PreprocessorDefinitions>INCLUDE_ELGATO;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+  </ItemDefinitionGroup>
+  <Target Name="Test" BeforeTargets="PrepareForBuild">
```
andreiborodin: I would name this target something a bit more verbose, like "TestForElgatoAndBlackMagic

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
```diff
@@ -128,27 +130,47 @@ private void CompositeGUI()
...
-                    if (fullscreenWindow != null)
+                    EditorGUILayout.Space();
+                    EditorGUILayout.BeginHorizontal();
                     {
-                        fullscreenWindow.TextureRenderMode = textureRenderMode;
+                        if (compositionManager != null)
+                        {
+                            GUI.enabled = compositionManager == null || !compositionManager.IsVideoFrameProviderInitialized;
+                            GUIContent label = new GUIContent("Video source", "The video capture card you want to use as input for compositing.");
+                            compositionManager.CaptureDevice = (FrameProviderDeviceType)EditorGUILayout.Popup(label, ((int)compositionManager.CaptureDevice), Enum.GetValues(typeof(FrameProviderDeviceType)).Cast<FrameProviderDeviceType>().Where(provider => compositionManager.IsFrameProviderSupported(provider) || (provider == FrameProviderDeviceType.None)).Select(x => x.ToString()).ToArray());
```
andreiborodin: Could you break this out into several lines variables; hard to read.

Also, I like to stack my linq-esk calls:
SomeList()
  .Where()
  .Select()
  .ToArray()

tools/ci/scripts/dependencies.props
```diff
@@ -9,9 +9,20 @@
...
+      <PreprocessorDefinitions>INCLUDE_BLACKMAGIC;INCLUDE_ELGATO;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+    <ClCompile Condition="Exists('$(DeckLink_inc)') And !Exists('$(Elgato_Filter)')">
+      <PreprocessorDefinitions>INCLUDE_BLACKMAGIC;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+    <ClCompile Condition="!Exists('$(DeckLink_inc)') And Exists('$(Elgato_Filter)')">
+      <PreprocessorDefinitions>INCLUDE_ELGATO;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+  </ItemDefinitionGroup>
+  <Target Name="Test" BeforeTargets="PrepareForBuild">
```
andreiborodin: Same as above, consider naming something more verbose.

src/SpectatorView.Native/SpectatorView.Compositor/UnityPlugin/UnityCompositorInterface.cpp
chrisfromwork: FWIW, The unity editor never unloads plugins so the clean up occurs when the program tears down. I'll see if I can find some sort of plugin unload call for this dll, but I'm not sure if it exists.

tools/ci/scripts/dependencies.props
chrisfromwork: sounds good.

src/SpectatorView.Native/SpectatorView.Compositor/dependencies.props
chrisfromwork: sounds good.

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
chrisfromwork: yeah that would help, let me fix things.

src/SpectatorView.Native/SpectatorView.Compositor/UnityPlugin/UnityCompositorInterface.cpp
chrisfromwork: looks like we have  unityPluginUnload call though which seems reasonable for cleanup

chrisfromwork: I've added a delete to the unload call.

PR #311, opened 2020-01-03
Remove text mesh pro preprocessor directive

chrisfromwork:
Having different compiled binaries that support text mesh pro vs don't support text mesh pro seems like overkill. For now, we can remove the text mesh pro pre processor directive to ease our compiled unity binary setup process. This should make our automation/nuget setup simpler for binary distribution of spectator view logic.

### DIFF
--- a/samples/Build2019Demo.Unity/ProjectSettings/ProjectSettings.asset
+++ b/samples/Build2019Demo.Unity/ProjectSettings/ProjectSettings.asset
@@ -627,21 +627,24 @@ PlayerSettings:
   webGLLinkerTarget: 1
   webGLThreadsSupport: 0
   scriptingDefineSymbols:
-    1: SPATIALALIGNMENT_ASA;STATESYNC_TEXTMESHPRO
-    4: SPATIALALIGNMENT_ASA;STATESYNC_TEXTMESHPRO
-    7: SPATIALALIGNMENT_ASA;STATESYNC_TEXTMESHPRO
-    14: SPATIALALIGNMENT_ASA;STATESYNC_TEXTMESHPRO
+    1: SPATIALALIGNMENT_ASA
+    4: SPATIALALIGNMENT_ASA
+    7: SPATIALALIGNMENT_ASA
+    14: SPATIALALIGNMENT_ASA
   platformArchitecture:
     iOS: 1
   scriptingBackend:
-    Metro: 2
+    Android: 1
+    Metro: 1
+    Standalone: 1
   il2cppCompilerConfiguration: {}
   managedStrippingLevel: {}
   incrementalIl2cppBuild: {}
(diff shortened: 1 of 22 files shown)


samples/Build2019Demo.Unity/ProjectSettings/ProjectSettings.asset
```diff
@@ -627,21 +627,24 @@ PlayerSettings:
   webGLLinkerTarget: 1
   webGLThreadsSupport: 0
   scriptingDefineSymbols:
-    1: SPATIALALIGNMENT_ASA;STATESYNC_TEXTMESHPRO
```
fibann: Are there any drawbacks with compiling the TextMeshPro bits unconditionally?

chrisfromwork: 1) It introduces a dependency on the text mesh pro unity package.
2) Folks consuming a different version of text mesh pro may have problems.

It's a bit unclear right now who uses Text Mesh Pro vs who doesn't. I think a more viable solution would be to move these classes into a separate dll with the text mesh pro dependency if folks voice that they really can't consume the Text Mesh Pro dependency.

I also think that the regressions caused by having these preprocessor definitions and the lack of discoverability around enabling/disabling these preprocessor definitions advocates for not keeping them.

chrisfromwork: technically right now the spectator view asmdef already has a text mesh pro dependency, so this preprocessor directive isn't helping avoid a text mesh pro dependency.

src/SpectatorView.Unity/Assets/SpectatorView.Tests/Microsoft.MixedReality.SpectatorView.Tests.msb4u.csproj.meta
```diff
@@ -0,0 +1,12 @@
...
+ScriptedImporter:
+  fileIDToRecycleName:
+    11400000: Microsoft.MixedReality.SpectatorView.Tests.msb4u
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
+  script: {fileID: 11500000, guid: 0e4dca1acba66fd478b3854f09a32ec7, type: 3}
+  buildEngine: 0
+  profiles: []
```
matthejo: Seems like these might be the same .meta files that probably should be excluded by .gitignore?

chrisfromwork: I'll ask andrei about this and will remove them as needed in a follow up review.

PR #316, opened 2020-01-13
Upgrade the SpectatorView.Native solution for VS 2019

matthejo:
This change upgrades the toolset version from v141 to v142 for all of the native projects. Doing so necessitated one code change, as _WCHAR_T_SIZE does not seem to be defined in the same include hierarchy it used to be.


chrisfromwork:
Tagging as a breaking change so that we have to write migration instructions.

PR #317, opened 2020-01-13
Move to Unity 2019.2.17f1 support and ASA v2.1.1

chrisfromwork:
This review contains the following changes:
1. Projects have been migrated to Unity 2019.2.17f1
1. Submodule/external download dependencies on ARCore and ARKit have been removed. Moving forward we will consume ARCore/ARKit/ARFoundation through Unity package manager.
1. Azure Spatial Anchors dependencies have been updated to v2.1.1 (Currently, we are using my personal fork because the azure repo does not contain some needed changes).
1. SpectatorView.Android and SpectatorView.iOS scenes have been replaced with SpectatorView.ARFoundation.Spectator
1. Various scripts have been rearranged in order to remove cyclic dependencies. Microsoft.MixedReality.SpatialAlignment.ASA no longer depends on Microsoft.MixedReality.SpectatorView. Microsoft.MixedReality.SpectatorView now depends on Microsoft.MixedReality.SpatialAlignment, Microsoft.MixedReality.SpatialAlignment.ASA, Microsoft.MixedReality.SpatialAlignment.ASA.Editor
1. SpatialAnchorsCoordinateService now utilizes CustomSpatialAnchorManager and CustomSpatialAnchorConfig (child classes based on ASA SDK code).
1. SpatialAnchorsCoordinate now needs to use the ARFoundation main camera to understand the permanent location of ASA coordinates. When applying transforms to the main camera, ASA coordinates move.
1. The SPATIALALIGNMENT_ASA preprocessor directive has been removed. Folks will now have to include azure spatial anchors in their project, but folks will also now not build projects with this configured incorrectly.
1. QR Code detection logic has preprocessor directives updated to avoid building code on mac. Extracting the QR Code dll zip fails on mac. (I will also be applying this change to master)

Notes for reviewing:
1. I would look primarily at new/updated *.cs files, there is a lot of noise with the Unity version change.

Outstanding items:
1. Samples should consume code through nuget packages/compiled dlls
1. We need to test android gradle and manifest definitions for the sample projects to make sure ASA still builds correctly for android/ios/hololens.
1. We need to update documentation/setup steps/setup scripts. This should come prior to attempting to merge this feature branch into master.

### Notes:
1. This change requires users to move to Unity 2019.2.17f1 and use ARFoundation compared to ARCore/ARKit. Unity logic for ASA v2.1.1 requires newer ARFoundation APIs not available in Unity 2018.
2. This change also automatically includes the ASA codebase in consumer projects, meaning, users will need to remove any preexisting symbolic links to ASA code from their project. 

### Migration Instructions:
1. Update your project to use Unity 2019.2.17f1.
2. Remove symbolic link references for ARKit-Unity-Plugin, AzureSpatialAnchors.Resources, AzureSpatialAnchorsPlugin, GoogleARCore and Plugins\Android from your project.


.gitmodules
```diff
@@ -3,13 +3,9 @@
...
 	branch = mrtk_development
 	ignore = dirty
-[submodule "external/ARCore-Unity-SDK"]
-	path = external/ARCore-Unity-SDK
-	url = https://github.com/google-ar/arcore-unity-sdk
-	ignore = dirty
 [submodule "external/Azure-Spatial-Anchors-Samples"]
 	path = external/Azure-Spatial-Anchors-Samples
-	url = https://github.com/Azure/azure-spatial-anchors-samples
+	url = https://github.com/chrisfromwork/azure-spatial-anchors-samples
```
wiwei: If we're going to take a dependency on this one (which is cool) instead of the official one I'm curious if there's anywhere that we can leave a comment/explain why this is necessary.

Doesn't seem like gitmodules has a comment capability though, so maybe this isn't doable that easily.

src/SpectatorView.Unity/Assets/PhotoCapture/Scripts/HoloLensCamera.cs
```diff
@@ -5,6 +5,8 @@
 // identity 'System.Numerics.Vectors, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' of 'System.Numerics.Vectors', you may need to supply runtime policy
 #pragma warning disable 1701
 
+#pragma warning disable CS0618 // 'WorldManager' is obsolete: 'Support for built-in VR will be removed in Unity 2020.1.
```
andreiborodin: Could you wrap the exact line (if not many) with this "#pragma warning disable" and "#pragma warning restore"

src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/AzureSpatialAnchors/CustomSpatialAnchorManager.cs
```diff
@@ -0,0 +1,48 @@
...
+using UnityEngine;
+
+namespace Microsoft.Azure.SpatialAnchors.Unity
+{
+    public static class CustomSpatialAnchorManagerExtensions
+    {
+        /// <summary>
+        /// Extension method that allows adding a SpatialAnchorManager with a custon SpatialAnchorConfig to a GameObject
+        /// </summary>
+        /// <param name="gameObject"></param>
```
wiwei: Nit: Don't have empty params or return statements (just delete them if you're not gonna add details). The generated docs look the same generally (i.e. using docfx or something else) so these don't add much value aside from padding the line count.

src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/AzureSpatialAnchors/CustomSpatialAnchorManager.cs
```diff
@@ -0,0 +1,48 @@
...
+// Licensed under the MIT License.
+
+using UnityEngine;
+
+namespace Microsoft.Azure.SpatialAnchors.Unity
+{
+    public static class CustomSpatialAnchorManagerExtensions
+    {
+        /// <summary>
+        /// Extension method that allows adding a SpatialAnchorManager with a custon SpatialAnchorConfig to a GameObject
```
wiwei:
```suggestion
        /// Extension method that allows adding a SpatialAnchorManager with a custom SpatialAnchorConfig to a GameObject
```

wiwei: Also, should this instead read "allows adding a CustomSpatialAnchorManager" instead of "SpatialAnchorManager?"

src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsCoordinate.cs
```diff
@@ -23,6 +26,26 @@ internal class SpatialAnchorsCoordinate : SpatialCoordinateUnityBase<string>
         /// <inheritdoc/>
         public override LocatedState State => LocatedState.Tracking;
 
+        private Matrix4x4 CoordinateTransform
+        {
+            get
+            {
+#if UNITY_ANDROID || UNITY_IOS
+                if (Camera.main == null)
```
wiwei: I think there have always been weird perf limitations assocaited with directly accessing Camera.main. If this is gonna be used a bunch (i.e. every frame) it might be worth caching Camera.main in a local var 

https://docs.unity3d.com/ScriptReference/Camera-main.html

'This property uses FindGameObjectsWithTag internally and doesn't cache the result. It is advised to cache the return value of Camera.main if it is used multiple times per frame.'

src/SpectatorView.Unity/MSBuild/settings.json
```diff
@@ -1 +1 @@
-{"autoGenerateEnabled":true}
\ No newline at end of file
+{"version":3,"autoGenerateEnabled":true,"dependenciesProjectGuid":"5a45860e-3555-4921-9cde-fb6d9945c4b8","assemblyCSharpGuid":"88a3e48c-339d-4416-92c3-307162279309","assemblyCSharpEditorGuid":"c9edf63e-d7b9-4f49-b5a0-4848a0205146","assemblyCSharpFirstPassGuid":"ccec8082-6754-4c7e-a742-8a94aa24b628","assemblyCSharpFirstPassEditorGuid":"2ac44e54-2ab1-47e5-b8e3-bd436d64e037","builtInPackagesFolderGuid":"2c55becf-70c1-4f9e-a90c-9e7f7688c9c4","importedPackagesFolderGuid":"91ab2fc0-d1d2-4bf1-8ce8-d160f619250b","externalPackagesFolderGuid":"3c35ba54-ff20-4034-a6e5-7696b24c5bf0","solutionGuid":"043e3c94-ee7e-405d-af1f-0fd50855f440"}
```
wiwei: Is it possible to prettify the json, or is this auto generated?

In general when checking in json/config, it's nicer to have them human readable (i.e. not just one line) so that things are also diffable going forward.

In the MRTK we had a massive json file that was just one line and it made diffing impossible. Then we just added new lines by running it through a formatter and it was a lot easier to deal with.

src/SpectatorView.Unity/Assets/PhotoCapture/Scripts/HoloLensCamera.cs
wiwei: It's generally nicer to do what @andreiborodin mentions because it keeps the warning ignores specific to the locations that need them (and thus makes each new warning explicitly annoying because don't want people adding more).

I think where this is fine though is if the entire file is also deprecated/non-sense without that dependency (i.e. if HoloLensCamera will be deleted in 2020) then it makes sense to have at the top level.

src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsCoordinate.cs
andreiborodin: Plus, it appears this check is only needed to be done once.

---
In reply to: [366091123](https://github.com/microsoft/MixedReality-SpectatorView/pull/317#discussion_r366091123) [](ancestors = 366091123)

src/SpectatorView.Unity/Assets/SpectatorView.Tests/Microsoft.MixedReality.SpectatorView.Tests.msb4u.csproj
```diff
@@ -21,6 +21,7 @@
   
   <Import Project="$([MSBuild]::GetPathOfFileAbove(MSBuildForUnity.Common.props))" Condition="Exists('$([MSBuild]::GetPathOfFileAbove(MSBuildForUnity.Common.props))')" />
   <PropertyGroup>
+    <DefineConstants>UNITY_INCLUDE_TESTS</DefineConstants>
```
andreiborodin: This should already be included in the generated props.

src/SpectatorView.Unity/MSBuild/settings.json
chrisfromwork: This doesn't seem like something for this review, it may be worth filing an issue against this in the MSBuildForUnity repo.

@andreiborodin what are your thougts?

src/SpectatorView.Unity/Assets/PhotoCapture/Scripts/HoloLensCamera.cs
chrisfromwork: I'm going to leave this how it is for now, these files will all have various issues in the future.

.gitmodules
chrisfromwork: We're going to do this for now to unblock the feature branch. I am working with the ASA team to get the changes we need into their codebase in parallel to this work.

src/SpectatorView.Unity/Assets/SpatialAlignment.ASA/Scripts/SpatialAlignment/SpatialAnchorsCoordinate.cs
chrisfromwork: Cached the camera and added some comments.

src/SpectatorView.Unity/MSBuild/settings.json
andreiborodin: Great idea! Please file an issue, or submit a PR ;)

tjsgtql:
i am happy to found this branch. but i run **SetupRepository.bat** have an error. it need a parameter for succeeded，i try 0,1,false,true,$false,$true,"false","1",,,, but it was not work

PR #320, opened 2020-01-23
Add Spectator View pre-build steps

chrisfromwork:
This change adds some spectator view pre-build steps:

1) Asset caches are automatically updated every build.
2) Updating asset caches every build can be disabled through a toolbar menu item.
3) Android manifest files/gradle files are dropped into a Assets/Plugins/Android folder every android build. Dialogs are shown to avoid automatically overwriting files when duplicate manifests are found.
4) The Spectator View screen recording activity is now automatically added to the ASA AndroidManifest file when building for android.
5) Pre-build steps not related to updating asset caches can also be disabled through a tool bar menu item.

NOTE: This review contains an MSBuildForUnity update unrelated to the changes. I would suggest looking exclusively at *.cs files.


src/SpectatorView.Unity/Assets/StreamingAssets/HiddenARCore.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
+guid: 90a1ed2970cd6564dbefade7e5ad0260
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant:
```
matthejo: This seems like it shouldn't be added, as I see no content in the PR for this folder

samples/SpectatorView.Example.Unity/.gitignore
```diff
@@ -72,4 +72,7 @@ UWP/
 
 # Avoid checking in build directories
 Assets/BuildTools.Unity
-Assets/BuildTools.Unity.meta
\ No newline at end of file
+Assets/BuildTools.Unity.meta
+
+# Avoid checking in asset caches
+Assets/Generated.StateSynchronization.AssetCaches/
```
matthejo: There are currently a few assets in this directory that aren't generated. We need to move those elsewhere.  I believe they are:

SpectatorViewSettings.prefab
StateSynchronizationSettings.prefab
DefaultStateSynchronizationPerformanceParameters.prefab

src/SpectatorView.Unity/Assets/StreamingAssets/HiddenARCore.meta
chrisfromwork: yeah sounds good

PR #325, opened 2020-02-24
Add color correction to the compositor for the camera stream

chrisfromwork:
Add the ability to apply color correction to the video camera feed. This review contains the following changes:

1) During OnPreRender, the TextureManager now has the ability to apply color correction to the video camera frame prior to starting compositing.
2) The CompositorWindow now contains UI that allows for users to perform color correction.

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
@@ -28,6 +28,7 @@ namespace Microsoft.MixedReality.SpectatorView.Editor
 
         private bool compositorStatsFoldout;
         private bool recordingFoldout;
+        private bool colorCorrectionFoldout;
         private bool hologramSettingsFoldout;
 
         private float hologramAlpha;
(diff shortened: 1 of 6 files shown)


src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
```diff
@@ -239,6 +241,120 @@ private void RecordingGUI()
...
+                        if (GUI.enabled &&
+                            !compositionManager.TextureManager.videoFeedColorCorrection.Enabled)
+                        {
+                            GUI.enabled = false;
+                        }
+
+                        EditorGUILayout.Space();
+                        EditorGUILayout.BeginHorizontal();
+                        {
+                            GUIContent label = new GUIContent("R Scale");
```
matthejo: >GUIContent label = new GUIContent("R Scale"); [](start = 28, length = 45)

Would be good to have tooltips that describe how these values all manipulate the color, or some documentation about how the ColorCorrection shader works with these values

chrisfromwork: testing these out locally, it seems like adding tool tips to these labels doesn't result in tool tips appearing in the compositor window. I will hold off on this for now.

PR #326, opened 2020-02-24
Add yield to OnPostRender for spectator view compositor

chrisfromwork:
If we don't wait for the end of the frame, post processing effects don't work correctly.


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/TextureManager.cs
```diff
@@ -302,8 +302,10 @@ private void OnPreRender()
             }
         }
 
-        private void OnPostRender()
+        private IEnumerator OnPostRender()
         {
+            yield return new WaitForEndOfFrame();
```
matthejo: >yield return new WaitForEndOfFrame(); [](start = 12, length = 37)

Can you add a comment about why this is here? It seems like it's to wait for other scripts that have OnPostRender steps to finish executing first

PR #327, opened 2020-02-25
Add support for the Azure Kinect camera

matthejo:
This commit contains a new feature: support for the compositor to use the Azure Kinect camera with either raw depth occlusion of holograms using the Kinect's depth camera, or depth+body tracking occlusion of holograms using the depth camera in conjunction with the Azure Kinect Body Tracking SDK.

Three new options are available, allowing you to use the Azure Kinect without occlusion, or to choose between the NFOV and WFOV modes on the depth camera.

image

When using the depth camera, you can choose to use only the raw depth images from the camera, or to use the body mask to only occlude people.

image

The depth camera image is compared with the Unity depth buffer for the camera to determine if pixels should be occluded or not.

image

The Azure Kinect camera uses the "Holographic Camera" user interface for detecting a marker to place the virtual camera. Currently only ArUco marker detection is supported.


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/CalibrationData.cs
```diff
@@ -1,102 +1,99 @@
...
-        /// <inheritdoc />
-        public void SetUnityCameraExtrinstics(Transform cameraTransform)
-        {
-            Matrix4x4 headFromCamera = cameraExtrinsics.ViewFromWorld;
-            cameraTransform.transform.localPosition = cameraExtrinsics.ViewFromWorld.GetColumn(3);
-            cameraTransform.transform.localRotation = Quaternion.LookRotation(cameraExtrinsics.ViewFromWorld.GetColumn(2), cameraExtrinsics.ViewFromWorld.GetColumn(1));
-
-            // Magic offset from Unity's underlying coordinate frame (WorldManager.GetNativeISpatialCoordinateSystemPtr()) and the head pose used for the camera.
-            // Poses are sent in the coordinate frame space because the Unity camera position uses prediction.
-            cameraTransform.localPosition += new Vector3(0f, 0.08f, 0.08f);
```
chrisfromwork: Did we determine this actually changed? What version of unity are we using?

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CalibrationDataProvider.cs
```diff
@@ -57,11 +57,23 @@ private async void SendCalibrationDataAsync()
...
                 {
                     byte[] contents = (await FileIO.ReadBufferAsync(file)).ToArray();
+                    if (CalculatedCameraCalibration.TryDeserialize(contents, out CalculatedCameraCalibration calibration))
+                    {
+                        // Magic offset from Unity's underlying coordinate frame (WorldManager.GetNativeISpatialCoordinateSystemPtr()) and the head pose used for the camera.
+                        // Poses are sent in the coordinate frame space because the Unity camera position uses prediction.
+                        Matrix4x4 viewFromWorld = calibration.Extrinsics.ViewFromWorld;
+                        Vector3 position = viewFromWorld.GetColumn(3);
+                        position += new Vector3(0f, 0.08f, 0.08f);
+                        viewFromWorld.SetColumn(3, position);
```
chrisfromwork: is there more context around why this is moved to here?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/MarkerDetection/ArUco/HoloLens/ArUcoMarkerDetector.cs
```diff
@@ -116,11 +122,28 @@ private void Update()
         {
             if (_detecting)
             {
+#if UNITY_EDITOR
+                int markerCount = UnityCompositorInterface.GetLatestArUcoMarkerCount();
```
chrisfromwork: the legibility of this code seems to suffer based on the editor vs non editor logic calls. Is there a major motivation behind not making a different ar uco marker detector that is specific to the compositor and implements IMarkerDetector? It seems like we don't use any of the same native logic calls? Someone may try and use this in the editor unrelated to the compositor if they see these #if UNITY_EDITOR stubs.

src/SpectatorView.Native/SpectatorView.Compositor/SharedHeaders/CompositorShared.h
```diff
@@ -55,11 +55,15 @@
...
 #define FRAME_BPP_RGBA 4
+#define FRAME_BPP_DEPTH16 2
+#define FRAME_BPP_BODY8 1
 #define FRAME_BPP_YUV  2
 #define FRAME_BPP_NV12 1.5f
 
 // Color camera buffer size.
 #define FRAME_BUFSIZE_RGBA          (FRAME_WIDTH * FRAME_HEIGHT * FRAME_BPP_RGBA)
+#define FRAME_BUFSIZE_DEPTH16       (FRAME_WIDTH * FRAME_HEIGHT * FRAME_BPP_DEPTH16)
+#define FRAME_BUFSIZE_BODY8       (FRAME_WIDTH * FRAME_HEIGHT * FRAME_BPP_BODY8)
```
chrisfromwork: >   [](start = 32, length = 2)

nit: add spaces so these lines lineup correctly

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectFrameProvider.cpp
```diff
@@ -0,0 +1,374 @@
...
+    if (K4A_RESULT_SUCCEEDED != k4a_device_open(K4A_DEVICE_DEFAULT, &k4aDevice))
+    {
+        OutputDebugString(L"Failed to open AzureKinect device");
+        goto FailedExit;
+    }
+
+    config.color_format = K4A_IMAGE_FORMAT_COLOR_BGRA32;
+    config.color_resolution = K4A_COLOR_RESOLUTION_1080P;
+    config.depth_mode = depthCameraMode;
+    config.camera_fps = K4A_FRAMES_PER_SECOND_30;
```
chrisfromwork: >K4A_FRAMES_PER_SECOND_30 [](start = 24, length = 24)

We may want to keep this at 1080p but would we want this 30fps to be configurable?

chrisfromwork:
> 

Do we think this copy step is something we should add to the CopyPluginsToUnity.bat script?

---
Refers to: src/SpectatorView.Native/README.md:111 in abe03e5. [](commit_id = abe03e5ec60a80f6e559159b76b8b504f4790c15, deletion_comment = False)

chrisfromwork:
> 

Can we consolidate ArUcoMarkerDetectors so that we don't end up with multiple files?

---
Refers to: src/SpectatorView.Native/SpectatorView.Compositor/Compositor/ArUcoMarkerDetector.h:28 in abe03e5. [](commit_id = abe03e5ec60a80f6e559159b76b8b504f4790c15, deletion_comment = False)

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectFrameProvider.h
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
chrisfromwork: > All rights reserved. [](start = 39, length = 21)

 nit: we no longer need to include "All rights reserverd."

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectFrameProvider.cpp
```diff
@@ -0,0 +1,374 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
chrisfromwork: >All rights reserved. [](start = 40, length = 20)

 nit: we no longer need to include "All rights reserverd."

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/ArUcoMarkerDetector.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
```
chrisfromwork: nit add license header

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/ArUcoMarkerDetector.cpp
```diff
@@ -0,0 +1,149 @@
+#include "stdafx.h"
+
```
chrisfromwork: nit: add license header

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/CameraIntrinsics.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
```
chrisfromwork: nit: license header

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/IFrameProvider.h
```diff
@@ -11,10 +13,19 @@ class IFrameProvider
     {
         BlackMagic,
         Elgato,
+        AzureKinect_DepthCamera_Off,
+        AzureKinect_DepthCamera_NFOV,
```
chrisfromwork: >AzureKinect_DepthCamera_NFOV [](start = 8, length = 28)

do we want to expose binned and unbinned?

chrisfromwork: it kind of seems like we may want to expose rgb and depth as separate providers. Or I can imagine a world where you may want to use azure kinect + dslr. 

---
In reply to: [384184940](https://github.com/microsoft/MixedReality-SpectatorView/pull/327#discussion_r384184940) [](ancestors = 384184940)

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/IFrameProvider.h
```diff
@@ -39,4 +50,12 @@ class IFrameProvider
     virtual int GetCaptureFrameIndex() { return 0; }
     virtual int GetPixelChange(int frame) { return 0; }
     virtual int GetNumQueuedOutputFrames() { return 0; }
+    virtual bool IsCameraCalibrationInformationAvailable() { return false; }
+    virtual void GetCameraCalibrationInformation(CameraIntrinsics* calibration) {}
+
+    virtual bool IsArUcoMarkerDetectorSupported() { return false; }
```
chrisfromwork: >IsArUcoMarkerDetectorSupported [](start = 17, length = 30)

It seems like every rgb camera should theoretically be able to support aruco marker detection. Should we consider an abstract base class where this is impleemented?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/CompositionManager.cs
```diff
@@ -499,6 +536,12 @@ private void Update()
                         CurrentCompositeFrame = 0;
                         timeSynchronizer.Reset();
                         poseCache.Reset();
+
+                        if (UnityCompositorInterface.IsCameraCalibrationInformationAvailable())
+                        {
+                            stationaryCameraBroadcaster.SetActive(true);
+                            HolographicCameraObserver.Instance.ConnectTo("127.0.0.1");
+                        }
```
chrisfromwork: could this be achieved without creating a connection to localhost/through establishing some sort of artificial connection?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/StationaryCameraCalibrationDataProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        private void NetworkManagerConnected(INetworkConnection obj)
+        {
+            SendCalibrationData();
+        }
+
+#if UNITY_EDITOR && UNITY_WSA
+        private void SendCalibrationData()
+        {
+            UnityCompositorInterface.GetCameraCalibrationInformation(out CompositorCameraIntrinsics compositorIntrinsics);
```
chrisfromwork: >GetCameraCalibrationInformation [](start = 37, length = 31)

it seems like this should be a tryget or we should check if calibration information is available before making this call.

chrisfromwork:
> using System.Collections;

nit: all of these new cs files need license headers:

---
Refers to: src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/StationaryCameraBroadcaster.cs:1 in abe03e5. [](commit_id = abe03e5ec60a80f6e559159b76b8b504f4790c15, deletion_comment = False)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/StationaryCameraCalibrationDataProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+            networkManager.Connected -= NetworkManagerConnected;
+        }
+
+        private void NetworkManagerConnected(INetworkConnection obj)
+        {
+            SendCalibrationData();
+        }
+
+#if UNITY_EDITOR && UNITY_WSA
+        private void SendCalibrationData()
```
chrisfromwork: does the azure kinect sdk work with UNITY_WSA? do we have any scenarios where we would build a UNITY_WSA stationary camera calibration data provider?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/StationaryCameraPoseProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            if (currentConnection == null || !currentConnection.IsConnected)
+            {
+                return;
+            }
+
+            if (sharedCoordinateParticipant == null)
+            {
+                SpatialCoordinateSystemManager.Instance.TryGetSpatialCoordinateSystemParticipant(currentConnection, out sharedCoordinateParticipant);
+            }
```
chrisfromwork: Does this mean that a stationary camera always reports its location as the location of the qr code/aruco marker in the scene?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemManager.cs
```diff
@@ -491,7 +493,7 @@ private bool TryCleanupExistingLocalizationSession(SpatialCoordinateSystemPartic
             {
                 if (currentLocalizationSession != null)
                 {
-                    if (currentLocalizationSession.Participant == participant)
+                    if (currentLocalizationSession.Participant == participant || currentLocalizationSession.CompletionSource.Task.IsCompleted)
```
chrisfromwork: >currentLocalizationSession [](start = 81, length = 26)

is there more context behind this? Could it be that we aren't setting currentLocalizationSEssion to null when we should be?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/PhysicalMarker/MarkerDetectorCoordinateService.cs
```diff
@@ -183,20 +183,30 @@ private async Task OnDiscoverCoordinatesImplAsync(CancellationToken cancellation
...
+#if UNITY_EDITOR
         private async Task OnDiscoverCoordinatesEditorAsync(CancellationToken cancellationToken, int[] idsToLocate = null)
         {
-            if (idsToLocate == null ||
-                idsToLocate.Length != 1)
+            if (UnityCompositorInterface.IsArUcoMarkerDetectorSupported())
             {
-                DebugLog("Running the MarkerDetectorCoordinateService in the editor only supports one coordinate id");
+                await OnDiscoverCoordinatesImplAsync(cancellationToken, idsToLocate);
                 return;
```
chrisfromwork: this UnityCompositorInterface call may need to be in a try catch if folks try and run mobile + hololens simulation in the editor.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/PhysicalMarker/MarkerDetectorCoordinateService.cs
```diff
@@ -183,20 +183,30 @@ private async Task OnDiscoverCoordinatesImplAsync(CancellationToken cancellation
             }
         }
 
+#if UNITY_EDITOR
         private async Task OnDiscoverCoordinatesEditorAsync(CancellationToken cancellationToken, int[] idsToLocate = null)
         {
-            if (idsToLocate == null ||
-                idsToLocate.Length != 1)
+            if (UnityCompositorInterface.IsArUcoMarkerDetectorSupported())
             {
```
chrisfromwork: do we need a try catch around UnityCompositorInterface calls? What happens for mobile simulation that executes in this code path if hte compositor dll doesn't exist locally?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/UnityCompositorInterface.cs
```diff
@@ -4,13 +4,68 @@
...
 using System.Runtime.InteropServices;
 using System.Text;
+using UnityEngine;
 
 namespace Microsoft.MixedReality.SpectatorView
 {
-    public enum FrameProviderDeviceType : int { BlackMagic = 0, Elgato = 1, None = 2 };
+    public enum FrameProviderDeviceType : int { BlackMagic = 0, Elgato = 1, AzureKinect_DepthCamera_Off = 2, AzureKinect_DepthCamera_NFOV = 3, AzureKinect_DepthCamera_WFOV = 4, None = 5 };
+    public enum OcclusionSetting : int { RawDepthCamera = 0, BodyTracking = 1};
     public enum VideoRecordingFrameLayout : int { Composite = 0, Quad = 1 };
```
chrisfromwork: it seems like we should have an option of none here

chrisfromwork: or at least it seems like we may want to support enabling an azure kinect depth frame provider + a black magic rgb frame provider at some point

---
In reply to: [384742662](https://github.com/microsoft/MixedReality-SpectatorView/pull/327#discussion_r384742662) [](ancestors = 384742662)

chrisfromwork:
it seems like this new marker detector can end up running indefinitely if an exception is thrown and the player is stopped

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectFrameProvider.cpp
```diff
@@ -0,0 +1,374 @@
...
+        auto height = k4a_image_get_height_pixels(image);
+        auto width = k4a_image_get_width_pixels(image);
+        auto buffer = k4a_image_get_buffer(image);
+
+        float focalLength[2] = { calibration.color_camera_calibration.intrinsics.parameters.param.fx, calibration.color_camera_calibration.intrinsics.parameters.param.fy };
+        float principalPoint[2] = { calibration.color_camera_calibration.intrinsics.parameters.param.cx, calibration.color_camera_calibration.intrinsics.parameters.param.cy };
+        float radialDistortion[3] = { calibration.color_camera_calibration.intrinsics.parameters.param.k1, calibration.color_camera_calibration.intrinsics.parameters.param.k2, calibration.color_camera_calibration.intrinsics.parameters.param.k3 };
+        float tangentialDistortion[2] = { calibration.color_camera_calibration.intrinsics.parameters.param.p1, calibration.color_camera_calibration.intrinsics.parameters.param.p2 };
+        markerDetector->DetectMarkers(buffer, width, height, focalLength, principalPoint, radialDistortion, tangentialDistortion, markerSize, cv::aruco::DICT_6X6_250);
+    }
```
chrisfromwork: we should expose this dictionary to the user

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/TextureManager.cs
```diff
@@ -329,10 +367,21 @@ private void OnPostRender()
...
+                Graphics.Blit(sourceTexture, blurOcclusionTexture, blurMat);
+
                 // Render the real-world video back onto the composited frame to reduce the opacity
                 // of the hologram by the appropriate amount.
                 holoAlphaMat.SetTexture("_FrontTex", renderTexture);
+                holoAlphaMat.SetTexture("_OcclusionTexture", blurOcclusionTexture);
                 Graphics.Blit(sourceTexture, compositeTexture, holoAlphaMat);
+
+                //var color = depthTexture.GetPixel(900, 400);
+                //Debug.Log(color);
```
chrisfromwork: nit: remove

chrisfromwork:
we need to parent Compositor_SharedSpatialCoordinate to its creator, when switching scenes this coordinate gets destroyed breaking the experience

chrisfromwork:
it seems like we may also need to talk through how to record occlusion masks when quadrant recording is enabled.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/CalibrationData.cs
matthejo: I moved this from the editor side to the HoloLens side. That way the offset of the HoloLens isn't baked into every CalibrationData, so CalibrationData can be reused between HoloLens+DSLR and AzureKinect.  Now this offset is applied in the HolographiCamera app before it sends the calibration data back to the compositor.

---
In reply to: [384166417](https://github.com/microsoft/MixedReality-SpectatorView/pull/327#discussion_r384166417) [](ancestors = 384166417)

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CalibrationDataProvider.cs
matthejo: Because CalibrationData is also now used for AzureKinect calibration (which doesn't have a hardcoded offset like HoloLens).

---
In reply to: [384168184](https://github.com/microsoft/MixedReality-SpectatorView/pull/327#discussion_r384168184) [](ancestors = 384168184)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/StationaryCameraCalibrationDataProvider.cs
matthejo: Really this should only happen in editor now, the UNITY_WSA check is probably not necessary. I'll remove it and also log a "Platform not supported" error message in the other SendCalibrationData block.

---
In reply to: [384190170](https://github.com/microsoft/MixedReality-SpectatorView/pull/327#discussion_r384190170) [](ancestors = 384190170)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/SpatialCoordinateSystemManager.cs
matthejo: I'll add a comment that this is needed for the stationary localization + world anchor when they're running in the same session

---
In reply to: [384193089](https://github.com/microsoft/MixedReality-SpectatorView/pull/327#discussion_r384193089) [](ancestors = 384193089)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/UnityCompositorInterface.cs
matthejo: I think in the future if we support combining different RGB and depth camera sources, that would make sense. In the current system, you're opting into occlusion (or not) solely through the choice of a FrameProviderDeviceType.

---
In reply to: [384743198](https://github.com/microsoft/MixedReality-SpectatorView/pull/327#discussion_r384743198) [](ancestors = 384743198,384742662)

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectFrameProvider.cpp
matthejo: 30 fps is the maximum supported, and I don't think we want to go lower than that

---
In reply to: [384179632](https://github.com/microsoft/MixedReality-SpectatorView/pull/327#discussion_r384179632) [](ancestors = 384179632)

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/IFrameProvider.h
matthejo: I didn't notice a quality difference in binned versus unbinned for NVOF, and unbinned WFOV is not supported at 30fps.

---
In reply to: [384185353](https://github.com/microsoft/MixedReality-SpectatorView/pull/327#discussion_r384185353) [](ancestors = 384185353,384184940)

chrisfromwork:
it seems like the elgato frame provider may not build correctly now with these changes

matthejo:
When you stop playback, StopFrameProvider is called from ResetCompositor, which destroys the AzureKinectFrameProvider. A new one would then be created upon restarting. So I don't believe the detector runs indefinitely across sessions. It will run until it detects a marker if you say "Locate Shared Spatial Coordinate", but that is equivalent to the behavior of the other providers.

---
In reply to: [591674762](https://github.com/microsoft/MixedReality-SpectatorView/pull/327#issuecomment-591674762) [](ancestors = 591674762)

PR #330, opened 2020-03-05
Monolithic move to start using upms

chrisfromwork:
This review does the following:

1. SpectatorView.Unity has been converted from a Unity project to a Unity package
2. SpectatorView.Example.Unity and Build2019.Demo now consume the com.microsoft.mixedreality.spectatorview package compared to contain symbolic links.

Notes for reviewing: I would not pay any attention to file moves. Some files did not get labeled by git as moves correctly, There are no logic changes in this review. The only items affected in the examples are manifest.json files that now reference the com.microsoft.mixedreality.spectatorview package.


chrisfromwork:
we will redo this to avoid the renames

PR #331, opened 2020-03-06
Update samples to consume code through unity packages

chrisfromwork:
this review contains the following:

1. SpectatorView.Unity is converted to a unity package
2. SpectatorView.Example.Unity and Build2019.Demo consume spectator view code now through the unity package
3. .gitignore and settings have been fixed for the example projects.


src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/CalibrationDataProvider.cs
```diff
@@ -60,7 +60,7 @@ private async void SendCalibrationDataAsync()
                     message.Write("CalibrationData");
                     message.Write(contents.Length);
                     message.Write(contents);
-                    networkManager.Broadcast(memoryStream.ToArray());
+                    networkManager.Broadcast(memoryStream.GetBuffer(), 0, memoryStream.Position);
```
matthejo: >networkManager.Broadcast(memoryStream.GetBuffer(), 0, memoryStream.Position); [](start = 20, length = 77)

Can we use TryGetBuffer here? that will still work in .NET Scripting. The Azure Kinect changed all of the GetBuffer usages to TryGetBuffer elsewhere

PR #354, opened 2020-03-27
Feature/unity2019

chrisfromwork:
Merging Unity 2019 feature branch into master. This review contains the following:

1) Spectator view source and dependencies now exist in a unity package that can be added to package manifest.json files.
2) External dependencies such as Azure Spatial Anchors and QR Code detection are moved to current releases.
3) Spectator View now builds in Unity 2019.
4) Android and iOS tracking is now based on AR Foundation. This was needed in order to use newer version of Azure Spatial Anchors.
5) Additional setup scripts have been added for compiling spectator view native dependencies via command line. It should only require one script to build and distribute a spectator view unity package amongst teams.


src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/StateSynchronizationMenuItems.cs
```diff
@@ -7,13 +7,42 @@
...
+                return PlayerPrefs.GetInt(disableSpectatorViewPreBuild, 0) > 0;
+            }
+            private set
+            {
+                PlayerPrefs.SetInt(disableSpectatorViewPreBuild, value ? 1 : 0);
+                PlayerPrefs.Save();
+            }
+        }
+
+        private const string disableUpdatingMenuItem = "Spectator View/Disable updating asset caches when building";
```
matthejo: >Disable updating asset caches when building [](start = 71, length = 43)

Menu items should be in Title Case, so should be "Disable Updating Asset Caches When Building"

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/StateSynchronizationMenuItems.cs
```diff
@@ -7,13 +7,42 @@
...
+            get
+            {
+                return PlayerPrefs.GetInt(disableUpdatingMenuItem, 0) > 0;
+            }
+            private set
+            {
+                PlayerPrefs.SetInt(disableUpdatingMenuItem, value ? 1 : 0);
+                PlayerPrefs.Save();
+            }
+        }
```
matthejo: I would prefer if these were both stated in the positive (Enable Spectator View Pre-Build Steps, Enable Updating Asset Caches When Building) and then have them be true by default. I find combining checked state with negatively-termed options to be kind of confusing

PR #355, opened 2020-04-02
Fix two K4A bugs related to body tracking.

hermann-noll:
This PR fixes:

- the compilation of the native compositor library when the kinect for azure is used without the body tracking SDK
- a crash when body tracking was not used but built as debug dll. <br/>some friendly entity initialized the `k4abtTracker` to `nullptr` when compiled as release. In debug mode msvc sets uninitialized members to `0xcd` so Unity crashed at `AzureKinectFrameProvider.cpp:348`.


chrisfromwork:
Thanks for this submission!

PR #356, opened 2020-04-02
Make the Azure Kinect provider use background threads to create a buffer of incoming video frames

matthejo:
There are two primary reasons for this change:
1. The Azure Kinect frame provider didn't correctly follow the convention of buffering input frames, and instead ignored the frame index the compositor requested and always rendered the current frame. This doesn't provide a guard against frame hitches.
2. The Azure Kinect frame provider blocked on the body tracking processing, which happens asynchronously on the GPU. This reduced the effective video framerate to around 20fps.

This change creates background threads to read frames from the camera and write them into a circular buffer. Those buffered changes are then consumed by the compositor.


src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectFrameProvider.h
```diff
@@ -23,65 +20,41 @@ class AzureKinectFrameProvider : public IFrameProvider
...
+        return cameraInput == nullptr ? 0 : cameraInput->GetCaptureFrameIndex();
     }
    
    virtual bool IsCameraCalibrationInformationAvailable() override
     {
         return true;
     }
 
-    virtual void GetCameraCalibrationInformation(CameraIntrinsics* calibration) override;
+   virtual void GetCameraCalibrationInformation(CameraIntrinsics* calibration) override { cameraInput->GetCameraCalibrationInformation(calibration); }
```
chrisfromwork, line 39: >cameraInput [](start = 90, length = 11)

nit: not sure when this is called but we should probably make sure that this camera input isn't nullptr

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectCameraInput.cpp
```diff
@@ -0,0 +1,401 @@
...
+            continue;
+        }
+
+        k4a_capture_t capture = nullptr;
+
+        switch (k4a_device_get_capture(_k4aDevice, &capture, K4A_WAIT_INFINITE))
+        {
+        case K4A_WAIT_RESULT_TIMEOUT:
+            OutputDebugString(L"Error: Timed out waiting for AzureKinect capture");
+            return;
```
chrisfromwork, line 180: >return [](start = 12, length = 6)

nit: it seems like a timeout may warrant a continue compared to an end of the update loop?

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectCameraFrame.cpp
```diff
@@ -0,0 +1,116 @@
...
+AzureKinectCameraFrame::~AzureKinectCameraFrame()
+{
+    for (int i = 0; i < AZURE_KINECT_IMAGE_TYPE_COUNT; i++)
+    {
+        delete[] _images[i];
+    }
+}
+
+void AzureKinectCameraFrame::StageImage(AzureKinectImageType imageType, k4a_image_t image)
+{
```
chrisfromwork, line 31: it seems like there should be some sort of guard around this. Like we should obligate the caller to declare it has the held lock

chrisfromwork: granted, it may just be better to have a try stage where we don't write if the status isn't writing

---
In reply to: [403176779](https://github.com/microsoft/MixedReality-SpectatorView/pull/356#discussion_r403176779) [](ancestors = 403176779)

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectCameraFrame.cpp
```diff
@@ -0,0 +1,116 @@
...
+void AzureKinectCameraFrame::EndWritingColorAndDepth()
+{
+    std::lock_guard<std::mutex> guard(_statusGuard);
+    if (_status == FrameStatus::WritingColorAndDepth)
+    {
+        _status = FrameStatus::Staged;
+    }
+}
+
+void AzureKinectCameraFrame::BeginWritingBodyMask()
```
chrisfromwork, line 75: >BeginWritingBodyMask [](start = 29, length = 20)

is there a reason that we have separate status for writing body mask vs writing color and depth?

chrisfromwork: It seems like we always write body mask while already in the color and depth mode, so there may not be a reason to have this additional status?

---
In reply to: [403178909](https://github.com/microsoft/MixedReality-SpectatorView/pull/356#discussion_r403178909) [](ancestors = 403178909)

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectCameraInput.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+void AzureKinectCameraInput::SetTransformedBodyMaskBuffer(uint16_t* transformedBodyMaskBuffer, int bufferSize)
+{
+    for (int i = 0; i < bufferSize; i++)
+    {
+        if (transformedBodyMaskBuffer[i] > 0)
+        {
+            transformedBodyMaskBuffer[i] = 1;
+        }
+    }
```
chrisfromwork, line 398: could this be done with a memset compared a loop?

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectCameraInput.cpp
```diff
@@ -0,0 +1,401 @@
...
+        {
+            bodyMaskBuffer[i] = 0;
+        }
+    }
+}
+
+void AzureKinectCameraInput::SetTransformedBodyMaskBuffer(uint16_t* transformedBodyMaskBuffer, int bufferSize)
+{
+    for (int i = 0; i < bufferSize; i++)
+    {
```
chrisfromwork, line 393: do we need to do this here or could it be done in a shader somewhere else?

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectCameraInput.cpp
```diff
@@ -0,0 +1,401 @@
...
+    while (!_stopRequested)
+    {
+        int frameIndex = _currentFrameIndex % MAX_NUM_CACHED_BUFFERS;
+        if (!_cameraFrames[frameIndex]->TryBeginWritingColorAndDepth())
+        {
+            // If the next frame in the buffer is still pending write or is being read from, then we've
+            // exceeded the capacity of the buffer and either the reader is blocking too long, or
+            // the body index processing thread has fallen behind.
+            OutputDebugString(L"Warning: frame buffer is completely full, and we can't begin writing to the next frame");
+            continue;
```
chrisfromwork, line 171: do we think that the color/depth capture should have authority here? Or is having the body tracking more important than keeping the color frame up to date?

chrisfromwork: another solution here seems to be to clear the body tracking data here compared to dropping the color/depth frame. I'm not sure which is worse.

---
In reply to: [403197138](https://github.com/microsoft/MixedReality-SpectatorView/pull/356#discussion_r403197138) [](ancestors = 403197138)

matthejo: I think that leaving the current behavior for now is ok, until we hit a case where the body tracking fails to keep up with the RGB camera. The result currently will be that the camera might fall below 30 frames per second, at which point the recorded video might have slight time slowdown.

---
In reply to: [403198425](https://github.com/microsoft/MixedReality-SpectatorView/pull/356#discussion_r403198425) [](ancestors = 403198425,403197138)

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/AzureKinectCameraInput.cpp
```diff
@@ -0,0 +1,401 @@
...
+            k4a_image_release(colorImage);
+        }
+        k4a_capture_release(capture);
+
+        // End the writing state. If the state machine already transitioned
+        // to WritingBodyMask, this will have no effect. If the state machine
+        // is still in the WritingColorAndDepth state, it will move to the 
+        // Staged state to mark the frame as ready for reading.
+        _cameraFrames[frameIndex]->EndWritingColorAndDepth();
+        _currentFrameIndex++;
```
chrisfromwork, line 230: do we want any sort of wait here or do we just want to keep looping with no pause?

chrisfromwork:
Do we know how safe it is to capture/transform frames off the ui thread?

PR #357, opened 2020-04-03
Feature: Use alternative capture card output

hermann-noll:
pr_alt_cc_output

## Description

My team and I are quite impressed by this project and plan to use it on various occasions, both video recording and live events. For the latter, as a matter of precaution, we would like to pass every output through a capture card (instead of connecting the computer directly to the stage beamer). This PR adds the ability to select an alternative video output for the compositor, so e.g. you can use a Kinect For Azure as input but output through a BlackMagic capture card.
I tested this feature with a BlackMagic Ultrastudio Express (as Half-Duplex) and an Ultrastudio 4K Mini (as Full-Duplex). Unfortunately I could not test with an Elgato device.

## Summary of changes

- `CompositorInterface` has a second, optional `IFrameProvider` pointer and related functions to set it / check the support for a given `ProviderType`
- The new GUI popup was added to `CompositorWindow` 
- `DeckLinkDevice` and `DeckLinkManager` had to modified to support output without capturing.
- `IFrameProvider::OutputYUV` was split into `ProvidesYUV` and `ExpectsYUV` to support cases where the input provider only supports BGRA (e.g. K4A), but the output provider expects YUV frames (e.g. default with BlackMagic)

## Further points of improvement

- If/When further output providers are added to this project it might be cleaner to separate the output capabilities into its own interface (e.g. `IFrameReceiver`)
- It might be helpful to have frame-provider-specific configuration in the compositor window (e.g. to set output display mode for BM, depth camera mode for K4A, etc.). This could also eliminate the need for a `USE_DECKLINK_SHUTTLE` macro.

### DIFF
--- a/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/CompositorInterface.cpp
+++ b/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/CompositorInterface.cpp
@@ -420,14 +420,22 @@ void CompositorInterface::RecordAudioFrameAsync(BYTE* audioFrame, LONGLONG audio
     activeVideoEncoder->QueueAudioFrame(audioFrame, audioSize, sampleTime);
 }
 
-bool CompositorInterface::OutputYUV()
+bool CompositorInterface::ProvidesYUV()
 {
-    if (outputFrameProvider != nullptr)
-        return outputFrameProvider->OutputYUV();
     if (frameProvider == nullptr)
     {
         return false;
     }
 
-    return frameProvider->OutputYUV();
+    return frameProvider->ProvidesYUV();
+}
+
+bool CompositorInterface::ExpectsYUV()
+{
+    if (outputFrameProvider != nullptr)
+        return outputFrameProvider->ExpectsYUV();
+    if (frameProvider != nullptr)
+        return frameProvider->ExpectsYUV();
+
+    return false;
 }
--- a/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/CompositorInterface.cpp
+++ b/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/CompositorInterface.cpp
@@ -19,6 +19,7 @@ CompositorInterface::CompositorInterface()
 
 void CompositorInterface::SetFrameProvider(IFrameProvider::ProviderType type)
 {
+    DisableOutputFrameProvider();
     if (frameProvider)
     {
         if (frameProvider->GetProviderType() == type)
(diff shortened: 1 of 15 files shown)


chrisfromwork:
We're excited to hear you are going to use the project! :)

FWIW, the elgato capture card that we currently support doesn't support an output stream, so there may not be a need to test here.

We'll look into this review Monday once we're back in the work week!

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/DeckLinkDevice.cpp
```diff
@@ -463,11 +463,13 @@ bool DeckLinkDevice::Init(ID3D11ShaderResourceView* colorSRV, ID3D11Texture2D* o
...
 
-    // Get input interface
-    if (m_deckLink->QueryInterface(IID_IDeckLinkInput, (void**)&m_deckLinkInput) != S_OK)
-        return false;
+        // Get input interface
+        if (m_deckLink->QueryInterface(IID_IDeckLinkInput, (void**)&m_deckLinkInput) != S_OK)
+            return false;
+    }
+    if (device == nullptr && outputTexture != nullptr)
+        outputTexture->GetDevice(&device);
```
matthejo, line 472: Can you add a comment on why this is necessary (e.g. when it will happen that either colorSRV is null or colorSRV fails to get a device)? It wasn't obvious to me reading the code.

matthejo: Ah, I think I understand, this happens for the output-only frame, because it doesn't have input.  In that case, I think it would be more clear for this second if to be an "else if"

---
In reply to: [404388384](https://github.com/microsoft/MixedReality-SpectatorView/pull/357#discussion_r404388384) [](ancestors = 404388384)

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/DeckLinkManager.cpp
```diff
@@ -86,7 +86,9 @@ HRESULT DeckLinkManager::Initialize(ID3D11ShaderResourceView* colorSRV, ID3D11Sh
                         videoDisplayMode = bmdMode4kDCI2398;
                     }
 
-                    deckLinkDevice->StartCapture(videoDisplayMode);
+                    if (colorSRV != nullptr)
+                        deckLinkDevice->StartCapture(videoDisplayMode);
```
chrisfromwork, line 90: nit: add parenthesis for legibility for this if statement

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/CompositorInterface.cpp
```diff
@@ -173,6 +218,8 @@ void CompositorInterface::UpdateFrameProvider()
     {
         frameProvider->Update(compositeFrameIndex);
     }
+    if (outputFrameProvider != nullptr)
+        outputFrameProvider->Update(compositeFrameIndex);
```
chrisfromwork, line 222: nit: add paranethsis around this if statement

chrisfromwork:
I added some really small nit critiques around code legibility, but in general I'd be happy to accept this change. Let us know if you can help with these comments. Otherwise, we can look into cleaning things up on our end.

hermann-noll:
For some reason the CLA bot did not check? An empty commit seems to have fixed it.

chrisfromwork:
@hermann-noll do you feel comfortable with us merging this into master?

hermann-noll:
Yes this PR should be good to go.

Issue #203, opened 2019-08-27
Support HoloLens 1 camera rigs


chrisfromwork:

Helco:
Hi, I am trying to wrap my head around this issue. I guess currently Hololens 1 isn't supported because it can't (natively) track QR Codes. But it should be possible to extrinsicly calibrate the Hololens to the Camera Image by the Aruco markers alone right? (tracked both by Hololens and the DSLR)
Is this just not implemented (yet) or is there some other problem I did not stumble upon?

Helco:
I have managed to get a first version of the calibration and the compositor working with the HL1. I printed the Calibration boards without the QR code markers.
I experienced some small issues regarding the detection of the Aruco markers for the extrinsic calibration (only a couple are detected at one given time, depending on the angle, full-frontal is at disadvantage). Also the holograms are swimming a bit around the place in the composited video.
Any thoughts regarding these issues?

Here are my relevant changes, ofc hacked together, but works as a PoC
```diff
diff --git a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Calibration/HeadsetCalibration.cs b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Calibration/HeadsetCalibration.cs
index a18e617..88621df 100644
--- a/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Calibration/HeadsetCalibration.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Calibration/HeadsetCalibration.cs
@@ -27,7 +27,8 @@ namespace Microsoft.MixedReality.SpectatorView
         /// </summary>
         [Tooltip("QR Code Marker Detector in scene")]
         [SerializeField]
-        protected QRCodeMarkerDetector qrCodeMarkerDetector;
+        protected GameObject markerDetectorGO;
+        private IMarkerDetector qrCodeMarkerDetector => markerDetectorGO.GetComponent<IMarkerDetector>();
 
         /// <summary>
         /// Debug Visual Helper in scene that will place game objects on qr code markers in the scene.
@@ -139,7 +140,11 @@ namespace Microsoft.MixedReality.SpectatorView
                     }
 
                     var originToQRCode = Matrix4x4.TRS(qrCodeTopLeftPosition, qrCodeRotation, Vector3.one);
[...]

PR #359, opened 2020-04-06
Feature: HL1 Camera Rigs

hermann-noll:
## Description

My team has started MR development with the HoloLens 1, now these devices are mostly lying around unused as the HL2 with its superior interactivity has taken over. For this reason I scribbled a version of a HL1 support for camera rigs in #203 (now a different account for internal compliance). I have cleaned and retested this approach and also got a correct spatial alignment working.

Unfortunately I could not compare the precision against a HoloLens 2, as I can't secure it properly to the camera at the moment (3D prints and matching screws are hard to come by rn).

## Summary of changes

- An `ArUcoMarkerDetector` was added to the `HeadsetCalibration` object and a condition whether to use QRCode or ArUco
- `ArUcoMarkerDetector` returns the static marker size in the `TryGetMarkerSize` which is contradictory to the docs ("was determined through detection"), but this function is afaik only used for calibration where it worked fine
- ` ArUcoThreeMarkerSpatialLocalizer` was enabled for the WSA platform.

## Calibration and Localization

The calibration was very finicky to get even somewhat working with the stationary marker detection strategy. Using the moving strategy and taking more pictures instead got plausible results for me (~30 usable datasets). For the test I moved the camera in a rough half circle around the calibration board. This I created by just stitching together all 18 ArUco codes from the original calibration board.
As far as I understood this method of calibration works because for the pose estimation the actual QR code positions are ignored and only the presumed ArUco code positions which are transferred by the HL2 in an original scenario are regarded in the calculation.

For the localization I found that the ArUco marker detection alone gives very unpredictable and inaccurate results. Using the three ArUco method instead to ignore the detected orientation and only rely on the detected positions gave a convincing result in the calibration test (the cube being half inside the marker is correct behaviour). For the situation in the picture below I had to put the rig very near to the marker board, presumably because I printed the markers too small (`0.036m`).
I had to switch the top and bottom marker IDs, I guess because in the usual use case the camera image is flipped?

|||
|:-:|:-:|


chrisfromwork:
This is great! :)

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Calibration/HeadsetCalibration.cs
```diff
@@ -243,5 +263,13 @@ private static MarkerCorners CalculateMarkerCorners(Vector3 topLeftPosition, Qua
             corners.orientation = topLeftOrientation;
             return corners;
         }
+
+#if UNITY_EDITOR
+        private void OnValidate()
+        {
+            FieldHelper.ValidateType<QRCodeMarkerDetector>(markerDetector);
+            FieldHelper.ValidateType<ArUcoMarkerDetector>(markerDetector);
```
chrisfromwork, line 271: I think the preference here would be to log an error on Enable if this type is wrong. Right now these two checks will succeed if the marker detector is null, which we probably dont' want.

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Calibration/HeadsetCalibration.cs
```diff
@@ -125,7 +139,7 @@ private void ProcessQRCodeUpdate()
             {
                 updatedMarkerIds.Add(marker.Key);
                 float size = 0;
-                if (qrCodeMarkerDetector.TryGetMarkerSize(marker.Key, out size))
+                if (markerDetector.TryGetMarkerSize(marker.Key, out size))
                 {
                     var qrCodeTopLeftPosition = CalcTopLeftFromCenter(marker.Value.Position, marker.Value.Rotation, size);
```
chrisfromwork, line 144: >qrCodeTopLeftPosition [](start = 24, length = 21)

we should probably change the naming here so that this isn't declared as a qr code. Although this works it will likely generate in the future.

chrisfromwork:
>                     markerPair.arucoMarkerCorners = CalculateMarkerCorners(arucoTopLeftPosition, arucoRotation, size);

Do we know whether the qr code is used for calibration/Should we avoid creating marker corners for the qr code when they are unavailable?

---
Refers to: src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Calibration/HeadsetCalibration.cs:177 in 41229ba. [](commit_id = 41229ba48a384cf16481510668bbbfa1d5ecf838, deletion_comment = False)

chrisfromwork:
I'd be more than happy to accept this change. I have made some comments around things that I think we should rename with this new support. If you can help address the comments great! If you would like help on our end let us know. :)

And thanks again for this submission!

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/SpatialAlignment/PhysicalMarker/ArUcoThreeMarkerSpatialLocalizer.cs
```diff
@@ -26,7 +26,7 @@ protected override bool IsSupported
         {
             get
             {
-#if UNITY_EDITOR
+#if UNITY_EDITOR || UNITY_WSA
```
matthejo, line 29: > || UNITY_WSA [](start = 16, length = 13)

If you're going to change this, please use the pattern that ArUcoMarkerSpatialLocalizer uses so that it's only supported on x86 UWP platform:

#if UNITY_EDITOR
                return true;
#elif UNITY_WSA
                return Windows.ApplicationModel.Package.Current.Id.Architecture == Windows.System.ProcessorArchitecture.X86;
#else
                return false;
#endif

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Calibration/HeadsetCalibration.cs
hermann-noll: I have renamed all related identifiers, except for those where a new name is required for the distinction between the two marker types (e.g. [debug visuals](https://github.com/microsoft/MixedReality-SpectatorView/pull/359/files?file-filters%5B%5D=.cs#diff-7a16755085586350abce224c5cfea0a0L53-L54)). Do you have a preference on such a naming?
A suggestion from my side would be `detectedMarker...` and `calculatedMarker...`

chrisfromwork: detectedMarker and calculatedMarker sounds good to me :)

chrisfromwork:
@hermann-noll do you feel comfortable with us merging this into master?

hermann-noll:
Yes this PR should be good to go.

Issue #314, opened 2020-01-10
Prevent device running the HolographicCamera application from going into standby


chrisfromwork:
https://docs.microsoft.com/en-us/uwp/api/Windows.System.Display.DisplayRequest

Calling DisplayRequest.RequestActive should prevent the device from going into standby while the HolographicCamera app is running.

PR #365, opened 2020-04-27
Keep camera mounted headset on when the holographic camera application is running

chrisfromwork:
This change fixes the following issue:
https://github.com/microsoft/MixedReality-SpectatorView/issues/314

We now have a monobehaviour/script in the HolographicCamera scene that makse sure the headset stays on while the application is running.

### DIFF
--- a/src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
+++ b/src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
@@ -8,67 +8,68 @@ using UnityEngine;
     using Windows.System.Display;
 #endif
 
-public class KeepDeviceActive : MonoBehaviour
+namespace Microsoft.MixedReality.SpectatorView
 {
-#if UNITY_WSA && WINDOWS_UWP
-    private static DisplayRequest displayRequest;
-#endif
-
-    private void Start()
-    {
-        CreateDisplayRequest();
-    }
-
-    private void OnDestroy()
+    /// <summary>
+    /// This class creates a UWP/WSA DisplayRequest in order to prevent Windows devices from entering inactive states.
+    /// For Holographic Camera, this functionality allows the HoloLens device to remain on when it is mounted to the camera.
+    /// Without this class, the HoloLens would enter an inactive state due to a lack of motion, which in turn can break filming.
+    /// </summary>
+    public class KeepDeviceActive : MonoBehaviour
     {
 #if UNITY_WSA && WINDOWS_UWP
-        if (displayRequest != null)
+        private static DisplayRequest displayRequest;
+
+        private void Start()
         {
-            try
-            {
-                displayRequest.RequestRelease();
-                Debug.Log("Display request released.");
-            }
-            catch (Exception e)
-            {
-                Debug.LogError($"Error releasing display request: {e.Message}");
-            }
+            CreateDisplayRequest();
         }
-#endif
-    }
 
-    private void CreateDisplayRequest()
-    {
-#if UNITY_WSA && WINDOWS_UWP
-        UnityEngine.WSA.Application.InvokeOnUIThread(() =>
+        private void OnDestroy()
         {
-            if (displayRequest == null)
+            UnityEngine.WSA.Application.InvokeOnUIThread(() =>
             {
-                try
+                if (displayRequest != null)
                 {
-                    displayRequest = new DisplayRequest();
-                }
-                catch (Exception e)
-                {
-                    Debug.LogError($"Error creating display request: {e.Message}");
-                }
-            }
+                    try
+                    {
+                        displayRequest.RequestRelease();
+                        Debug.Log("Display request released.");
+                    }
+                    catch (Exception e)
+                    {
+                        Debug.LogError($"Error releasing display request: {e.Message}");
+                    }
 
-            if (displayRequest != null)
+                    displayRequest = null;
+                }
+            }, true);
+        }
+
+        private void CreateDisplayRequest()
+        {
+            UnityEngine.WSA.Application.InvokeOnUIThread(() =>
             {
-                try
+                if (displayRequest == null)
                 {
-                        // This call activates a display-required request. If successful,  
-                        // the screen is guaranteed not to turn off automatically due to user inactivity. 
-                        displayRequest.RequestActive();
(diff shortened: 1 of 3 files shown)


src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
```diff
@@ -0,0 +1,74 @@
...
+// Licensed under the MIT License.
+
+using UnityEngine;
+
+#if UNITY_WSA && WINDOWS_UWP
+    using System;
+    using Windows.System.Display;
+#endif
+
+public class KeepDeviceActive : MonoBehaviour
```
matthejo, line 11: Nit: Enclose this in the Microsoft.MixedReality.SpectatorView namespace

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
```diff
@@ -0,0 +1,74 @@
...
+                }
+            }
+
+            if (displayRequest != null)
+            {
+                try
+                {
+                        // This call activates a display-required request. If successful,  
+                        // the screen is guaranteed not to turn off automatically due to user inactivity. 
+                        displayRequest.RequestActive();
```
matthejo, line 63: Nit: indentation is inconsistent

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (displayRequest == null)
+            {
+                try
+                {
+                    displayRequest = new DisplayRequest();
+                }
+                catch (Exception e)
+                {
+                    Debug.LogError($"Error creating display request: {e.Message}");
+                }
```
matthejo, line 54: Is it really possible for the construction of the DisplayRequest to throw an exception?

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            try
+            {
+                displayRequest.RequestRelease();
+                Debug.Log("Display request released.");
+            }
+            catch (Exception e)
+            {
+                Debug.LogError($"Error releasing display request: {e.Message}");
+            }
```
matthejo, line 35: This should probably set the displayRequest to null after releasing it.

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
```diff
@@ -0,0 +1,74 @@
...
+
+    private void Start()
+    {
+        CreateDisplayRequest();
+    }
+
+    private void OnDestroy()
+    {
+#if UNITY_WSA && WINDOWS_UWP
+        if (displayRequest != null)
```
matthejo, line 25: >f (displayRequest != null) [](start = 9, length = 26)

Does this need the InvokeOnUIThread wrapper as well?

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
wiwei, line 63: Thoughts on moving displayRequest.RequestActive() to the previous try block (i.e. right after creating it?)

I guess the downside is that you might not get specific Debug.LogError for creation vs requesting, but it seems a bit messy here to have a try/catch for each of these.

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
```diff
@@ -0,0 +1,74 @@
...
+            {
+                Debug.LogError($"Error releasing display request: {e.Message}");
+            }
+        }
+#endif
+    }
+
+    private void CreateDisplayRequest()
+    {
+#if UNITY_WSA && WINDOWS_UWP
```
wiwei, line 42: Is it possible to use this  #if wrapper at a higher level (i.e. completely enclosing Start/OnDestroy/CreateDisplayRequest?)

The micro-managey part about the current set up is a bit inconsistent in where things get enclosed (i.e. why is ifdefing caught within this function instead of inside start, is there a lot of value to having functions that end up just being empty on non UWP, etc)

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
wiwei, line 11: Add a doccomment explaining what this class does.

It's actually a super useful class in general given the API that it invokes.

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
chrisfromwork: I'm not sure why it would, but this mimics the sample behavior where it suggests that it can:
https://docs.microsoft.com/en-us/uwp/api/windows.system.display.displayrequest?view=winrt-18362

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
chrisfromwork: Yeah, seems like it should

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
chrisfromwork: I based this off of a sample, but reducing the length of this code seems like a good plan. I would hope that the exceptions would be different, I'm not really sure how to generate either of them.

src/HolographicCamera.Unity/Assets/HolographicCamera/Scripts/KeepDeviceActive.cs
chrisfromwork: sounds good

PR #369, opened 2020-05-08
Fix crash when k4a_device_open fails

hermann-noll:
When `k4a_device_open` fails, the `_cameraFrames` array was never initialized, so the Unity Editor crashes in the `UpdateSRVs` method. When an early-return is added there it just crashes in the destructor of `AzureKinectCameraFrame` where it is not as easy to add such an early-return.

Instead I moved the initialization of the array so it is always initialized, with this I could not reproduce this crash.


chrisfromwork:
Thanks for this fix!

PR #377, opened 2020-05-20
Feature: Integrated Device Portal

hermann-noll:
PR-DevicePortal

## Description

These days live demos are mostly done by me juggling many different programs simultaneously 
 (Unity, Hololens App, Teams, OBS, Browser) to finally be able to stream a video signal. One of these programs is the browser as to start/stop the holographic camera app on the Hololens.
With this PR I can now do this directly in the Unity editor removing the need for the browser. Also I added some basic health information so I can judge before/during the demo whether the app is about to crash (using CPU/RAM as indicators) and whether the HL is going to shutdown on me because I forgot to charge it sufficiently beforehand.

I tested this feature with both HL1 and HL2 

## Summary of changes

- A new GameObject with a single component `DevicePortalConnection` in the Compositor prefab that handles the web requests to the DevicePortal
- The UI was added to the `CompositorWindow`

### DIFF
--- a/src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
@@ -610,7 +610,7 @@ namespace Microsoft.MixedReality.SpectatorView.Editor
                     case DevicePortalState.Connecting:
                         stateText = "<color=silver>Connecting...</color>";
                         buttonText = "Cancel";
-                        onPressButton = devicePortal.Cancel;
+                        onPressButton = devicePortal.CancelConnect;
                         break;
 
                     case DevicePortalState.NotStarted:
@@ -635,7 +635,7 @@ namespace Microsoft.MixedReality.SpectatorView.Editor
                     case DevicePortalState.Starting:
                         stateText = "<color=orange>Starting...</color>";
                         buttonText = "Cancel";
-                        onPressButton = devicePortal.Cancel;
+                        canPressButton = false;
                         break;
 
                     case DevicePortalState.Stopping:
--- a/src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
+++ b/src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
@@ -670,6 +670,9 @@ namespace Microsoft.MixedReality.SpectatorView.Editor
                         }
                     }
 
+                    GUI.enabled = false;
+                    EditorGUILayout.TextField("IP address", holographicCameraIPAddress);
+
                     GUI.enabled = prevEnabled && devicePortal.State == DevicePortalState.NotConnected;
                     devicePortalUser = EditorGUILayout.TextField("Username", devicePortalUser);
                     devicePortalPassword = EditorGUILayout.PasswordField("Password", devicePortalPassword);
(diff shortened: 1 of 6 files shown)


src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/DevicePortalConnection.cs
```diff
@@ -0,0 +1,310 @@
+using System;
+using System.Linq;
```
chrisfromwork, line 2: nit: please add the following header to this file:

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/DevicePortalConnection.cs
```diff
@@ -0,0 +1,310 @@
...
+                return;
+            }
+            if (State != DevicePortalState.NotStarted)
+            {
+                Debug.LogError($"Unexpected state for starting the app: {State}");
+                return;
+            }
+            State = DevicePortalState.Starting;
+            var cts = new CancellationTokenSource();
+            Dispatcher.ScheduleAsync(StartAppAsync, cts.Token);
```
chrisfromwork, line 139: nit: if we aren't ever cancelling these CancellationTokenSources declared in line, it may be more legible to use CancellationToken.None instead

chrisfromwork:
In general this review looks great. I think everyone will appreciate having this change.

A few questions on the ux:
are we basing the connection attempt on the ip address specified for the holographic camera further up in the compositor window? If so, would it be possible to add a non-editable text line for that ip address displayed in the device portal drop down? When trying this out locally I wasn't 100% sure what ip address we were using for the device portal connection attempt. I think that reshowing the ip address in the device portal drop down could help. But, what are your thoughts?

Regardless, I think the main thing from preventing this getting checked in right now is the missing header for the DevicePortalConnection.cs file. Once that is resolved we can likely check this in.

hermann-noll:
I would guess that the device portal is mainly used with the HL on the camera rig (as you can't put it on easily), so I had the connection address fixed to this one. Stating the IP address in the dropdown which is used is definitely a good idea, I added it with the last commit. It updates automatically as you type the ip adress for the holographic camera.

image

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
```diff
@@ -573,6 +580,115 @@ private void CompositorStatsGUI()
...
+
+                    string stateText, buttonText;
+                    bool canPressButton = true;
+                    Action onPressButton = () => { };
+                    switch (devicePortal.State)
+                    {
+                        case DevicePortalState.NotConnected:
+                            stateText = "<color=silver>Not Connected</color>";
+                            canPressButton =
+                                devicePortalUser.Trim() != "" &&
```
matthejo, line 607: >devicePortalUser.Trim() [](start = 32, length = 23)

Nit: Could be a devicePortalUser.IsNullOrWhitespace call here

chrisfromwork:
I was able to test out this locally and greatly appreciate not having to open device portal.

Do you think it would be worth adding the ability to disconnect from a device via device portal? I don't know how often folks will be switching between holographic cameras or if they would ever do so in the same session. But only being able to start and stop the holographic camera app and not disconnect from device portal without stopping the scene might end up as a bug further down the road.

Regardless, this review looks good. Let me know when you feel its ready to check in and I will merge the PR. Thanks again for these contributions! :)

hermann-noll:
That is a good point, a possible use-case might be having multiple viewpoints with a single computer or being able to switch power-drained devices quickly without having to restart the app.
As for UX I opted for just an optional second button to the side, which I personally preferred over a dropdown menu with the two actions (e.g. Start App & Disconnect).

image

If you are okay with this style I think this PR is good to be merged.

chrisfromwork:
Sounds good. I'll look into merging this today. :)

PR #381, opened 2020-05-25
Feature: Add low latency mode

hermann-noll:
## Description

My team is exploring the usage of Steadycams for being mobile whilst using heavier production cameras, as these are difficult to carry around and move smoothly for the duration of the shoot. In this scenario the monitor on the rig can provider the operator with a live view with the holograms to be able to frame them properly.
The buffering of frames works great for smooth recordings but also introduces quite a bit of latency on top of the necessary technical latency by transmitting the video signal from and back to the camera rig. This PR adds a low-latency mode switch which reduces both the buffering done in the input ([`CompositionManager.cs:507`](https://github.com/microsoft/MixedReality-SpectatorView/pull/381/files#diff-a831c0726943bca9148b344f03d11ffbL507)) and in the DeskLink Output ([`DeckLinkDevice.cpp:246`](https://github.com/microsoft/MixedReality-SpectatorView/pull/381/files#diff-0c30655b993435b2c4d00acb52bfb693L246)).

In a simple test of using a stopwatch to measure the perceived latency between a visual change in front of the camera to it being seen on the rig monitor this achieves a drop from ~1,1 seconds down to ~0,4 seconds. Which still isn't great but a necessary improvement for these types of rigs.

## Summary of Changes

- A new property `CompositionManager.LowLatencyMode` which is used for initialization and for stepping forward in the input frames.
- To toggle this the `CompositorWindow` UI has been adapted
- A new frame provider method `SetOutputLowLatencyMode` is propagated through all the layers of the native compositor
- The `OutputScheduler` of the decklink frame provider uses a configurable limit to judge the health of the output queue

### DIFF
--- a/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/DeckLinkDevice.cpp
+++ b/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/DeckLinkDevice.cpp
@@ -288,9 +288,10 @@ public:
 
     }
 
-    void SetLowLatencyMode(bool isEnabled)
+    void SetLatencyPreference(float latencyPreference)
     {
-        minFramesQueued = isEnabled ? 1 : defaultMinFramesQueued;
+        minFramesQueued = 1 + latencyPreference * (minFramesQueued - 1);
+        minFramesQueued = min(1UL, max(minFramesQueued, static_cast<unsigned long>(MAX_NUM_OUTPUT_FRAMES)));
     }
 
     HRESULT	STDMETHODCALLTYPE ScheduledFrameCompleted(IDeckLinkVideoFrame* completedFrame, BMDOutputFrameCompletionResult result)
@@ -742,9 +743,9 @@ int DeckLinkDevice::GetNumQueuedOutputFrames()
     return s_outputScheduler.framesQueued;
 }
 
-void DeckLinkDevice::SetOutputLowLatencyMode(bool isEnabled)
+void DeckLinkDevice::SetLatencyPreference(float latencyPreference)
 {
-    s_outputScheduler.SetLowLatencyMode(isEnabled);
+    s_outputScheduler.SetLatencyPreference(latencyPreference);
 }
 
 void DeckLinkDevice::Update(int compositeFrameIndex)
--- a/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/DeckLinkDevice.cpp
+++ b/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/DeckLinkDevice.cpp
@@ -51,6 +51,8 @@ public:
 #define MAX_NUM_OUTPUT_FRAMES 20
     IDeckLinkMutableVideoFrame* outputVideoFrames[MAX_NUM_OUTPUT_FRAMES];
     int outputWriteIndex = 0;
+    const unsigned long defaultMinFramesQueued = MAX_NUM_OUTPUT_FRAMES - 4;
+    unsigned long minFramesQueued = defaultMinFramesQueued;
 
     OutputScheduler()
     {
@@ -219,7 +221,7 @@ public:
         if (started)
         {
             //Try to fill the queue back up
-            if (framesQueued < MAX_NUM_OUTPUT_FRAMES - 5)
+            if (framesQueued < minFramesQueued - 1)
                 duration += 20;
 
             //Make sure we dont fall behind on big hitches
@@ -243,7 +245,7 @@ public:
             if (!started)
             {
                 //Buffer frames, then start
-                if (framesQueued > MAX_NUM_OUTPUT_FRAMES - 4)
+                if (framesQueued >= minFramesQueued)
                 {
                     // Start
                     result = m_deckLinkOutput->StartScheduledPlayback(0, timeScale, 1.0);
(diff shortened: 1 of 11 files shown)


src/SpectatorView.Native/SpectatorView.Compositor/Compositor/DeckLinkDevice.cpp
```diff
@@ -286,6 +288,11 @@ class OutputScheduler : public IDeckLinkVideoOutputCallback
 
     }
 
+    void SetLowLatencyMode(bool isEnabled)
+    {
+        minFramesQueued = isEnabled ? 1 : defaultMinFramesQueued;
+    }
```
chrisfromwork, line 294: Would it be better take in a float or double here and interpolate between these two values with a slider? Or do we know that we will always want this toggle between 1 frame queued vs default frames queued?

src/SpectatorView.Unity/Assets/SpectatorView/Scripts/Compositor/CompositionManager.cs
```diff
@@ -504,7 +526,7 @@ private void Update()
 
             //Set our current frame towards the latest captured frame. Dont get too close to it, and dont fall too far behind 
             int step = (captureFrameIndex - CurrentCompositeFrame);
-            if (step < 8)
+            if (step < (LowLatencyMode ? 2 : 8))
```
chrisfromwork, line 529: >2 [](start = 41, length = 1)

nit: I know that these line values weren't your doing, but would it be possible to get these values defined as constants?

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
```diff
@@ -184,6 +184,13 @@ private void CompositeGUI()
 
                             compositionManager.OutputDevice = supportedOutputDevices[selectedIndex];
 
+                            EditorGUILayout.EndHorizontal();
+                            EditorGUILayout.Space();
+                            EditorGUILayout.BeginHorizontal();
+
+                            var lowLatencyLabel = new GUIContent("Low Latency Mode", "Disables buffers to prevent stutters on performance or IO problems. Enable only for live production.");
+                            compositionManager.LowLatencyMode = EditorGUILayout.Toggle(lowLatencyLabel, compositionManager.LowLatencyMode);
```
chrisfromwork, line 192: Does this only impact black magic output streams or does this also impact the frame shown in the compositor? If this only impacts black magic playback, it seems like we may want this property hidden behind IFrameProvider::SupportsOutput

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/DeckLinkDevice.cpp
hermann-noll: Personally I would probably only change between low-latency and high-safety but I can see that one might want to have finer control over this and it should be easy-enough to implement.

src/SpectatorView.Unity/Assets/SpectatorView.Editor/Scripts/CompositorWindow.cs
hermann-noll: The low latency mode also reduces the size of the incoming frame queue. 
If there is a probable use-case where the output via the preview window would need to be low-latency hiding the switch/slider might spark some confusion?

To prevent this property to be too prominent towards the selection of the frame providers we might also move it in the UI below "Frame time adjustment" where it might fit better.

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/DeckLinkDevice.cpp
```diff
@@ -286,6 +288,12 @@ class OutputScheduler : public IDeckLinkVideoOutputCallback
 
     }
 
+    void SetLatencyPreference(float latencyPreference)
+    {
+        minFramesQueued = 1 + latencyPreference * (minFramesQueued - 1);
```
matthejo, line 293: >minFramesQueued [](start = 51, length = 15)

Should this be defaultMinFramesQueued?  Otherwise it looks like each call to SetLatencyPreference will keep reducing the minFramesQueued without ever growing the value

matthejo: Thinking more, maybe actually I meant MAX_NUM_OUTPUT_FRAMES?

---
In reply to: [432713687](https://github.com/microsoft/MixedReality-SpectatorView/pull/381#discussion_r432713687) [](ancestors = 432713687)

hermann-noll: good catch!
`defaultMinFramesQueued` had a couple of frames left for mitigating delays by the decklink output, so using `MAX_NUM_OUTPUT_FRAMES` and a `latencyPreference` of `1` would disable this.

chrisfromwork:
I was able to test this out locally today and it seems to work well. :)

Let me know when you feel its ready to check in! And thanks again for these contributions.

hermann-noll:
Great to hear it does work on other machines as well!

Unfortunately I have encountered a bug where recorded videos sometimes jump as old frames are inserted. Of course this might be unrelated to this PR, but in case it is not I would like to investigate that one over the weekend.

chrisfromwork:
Let us know what you figure out. I think there are definitely some bugs floating around in the encoder stack where starting samples get messed up if the compositor has been running a long time. In my experience, I have been able to record a single video with success while I have seen errors in encoding multiple videos during the same player run prior to this change.

hermann-noll:
After spending some more time with the video encoder, I am somewhat sure that this PR does not create more stutters or messed up frames in recorded videos than before (not more than a slightly higher CPU usage does). Meaning this PR should be good to go.

I'll still look a bit more into the encoder stuff, as there are some points to optimize (e.g. the documented encoding by texture or removing unnecessary memory copies).

chrisfromwork:
sounds good. Thank you for this contribution!

PR #395, opened 2020-06-18
Improving Videoencoder

hermann-noll:
So after a couple of shooting days we had some issues with the resulting footage, namely some stutters by duplicated frames and what seemed like frames starting and ending at wrong times relative to the video (video pausing for up to several minutes at starting and freezing at the end). This is why I tried looking into improving the videoencoder a bit, ideally lowering the required CPU usage especially for split channels as these are very high leaving not much room for the application itself (jumping from ~ca 30% to 70-80% on a modern 6-core CPU).

I found a couple of things but as some of these changes are a bit bigger I would like to check with you whether you are OK with the proposed API changes or if you have further insights which I missed. Also I am sorry in beforehand for the long text, I deemed it necessary to explain the complexity involved.

### Unnecessary Memory allocations/copies
With the current state of master a rendered frame takes this way to the `IMFSinkWriter`: 
- Fetched from GPU into a (unguarded) ring of 10 buffers owned by UnityCompositorInterface ([UnityCompositorInterface.cpp:181](https://github.com/microsoft/MixedReality-SpectatorView/blob/7796da6acb0ae41bed1b9e0e9d1c5c683b4b8374/src/SpectatorView.Native/SpectatorView.Compositor/UnityPlugin/UnityCompositorInterface.cpp#L181))
- Copied into a newly allocated temporary buffer owned by the current videoencoder ([VideoEncoder.cpp:346](https://github.com/microsoft/MixedReality-SpectatorView/blob/7796da6acb0ae41bed1b9e0e9d1c5c683b4b8374/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp#L346))
- Copied into another newly allocated `IMFMediaBuffer` owned by the `WriteVideo` task ([VideoEncoder.cpp:387](https://github.com/microsoft/MixedReality-SpectatorView/blob/7796da6acb0ae41bed1b9e0e9d1c5c683b4b8374/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp#L387))
- (Presumably) uploaded to GPU for hardware encoding; alternatively processed directly by the CPU 

These copies as far as I can see are done for buffering (multiple render events for one `UpdateCompositor` call), multithreading (so the `WriteVideo` task does not access the ring pool) and maybe encapsulation (so the UnityCompositorInterface does not have to know about Media Foundation)? I drafted an alternative method to deal with these intentions without copies although with a bit looser encapsulation, based on move semantics. So (for CPU encoding) the new flow would look like this (probably best explored at 9cc0fbb):

- at initialization the videoencoder allocates a pool of 10 frame buffers ([VideoEncoder.cpp:66](https://github.com/hermann-noll/MixedReality-SpectatorView/blob/09f62de01c1419c20dd7e1a1b1ea3ce47f4db7b2/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp#L66))
- The UnityCompositorInterface gets a new frame buffer (by `std::unique_ptr`) from the videoencoder ([UnityCompositorInterface.cpp:125](https://github.com/hermann-noll/MixedReality-SpectatorView/blob/09f62de01c1419c20dd7e1a1b1ea3ce47f4db7b2/src/SpectatorView.Native/SpectatorView.Compositor/UnityPlugin/UnityCompositorInterface.cpp#L125))
- these frames own an lockable `IMFMediaBuffer` in which the texture is fetched ([VideoEncoder.h:127](https://github.com/hermann-noll/MixedReality-SpectatorView/blob/09f62de01c1419c20dd7e1a1b1ea3ce47f4db7b2/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.h#L127))
- the UnityCompositorInterface moves the frame back to the videoencoder with `QueueVideoFrame` ([UnityCompositorInterface.cpp:128](https://github.com/hermann-noll/MixedReality-SpectatorView/blob/09f62de01c1419c20dd7e1a1b1ea3ce47f4db7b2/src/SpectatorView.Native/SpectatorView.Compositor/UnityPlugin/UnityCompositorInterface.cpp#L128))
- the videoencoder calculates the right sample time and moves the frame into the asynchronous task of `WriteVideo` ([VideoEncoder.cpp:366](https://github.com/hermann-noll/MixedReality-SpectatorView/blob/09f62de01c1419c20dd7e1a1b1ea3ce47f4db7b2/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp#L366))
- when the task finishes it moves the frame back into the pool ([VideoEncoder.cpp:411](https://github.com/hermann-noll/MixedReality-SpectatorView/blob/09f62de01c1419c20dd7e1a1b1ea3ce47f4db7b2/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp#L411))

This way the fetched frame is never copied (at least not by the SpectatorView) around in memory and no temporary memory is allocated and freed per frame. The biggest caveat for me with this would be the exposing of `IMFMediaBuffer` although the UnityCompositorInterface does not have to interact with it.

For implementation I had to replace the usage of the `concurrency::create_task` from the PPL, also I could not use `concurrency::concurrent_queue` for the frame pool as the PPL does not seem to support move-only datatypes. Luckily there is a standard c++ functionality for the `create_task` (`std::async` with the `std::launch::async` policy set) and a simple pair of `std::queue` and `std::mutex` for the pool.

### Potentially unordered `WriteSample` calls

As far as I understood multiple `concurrency::create_task` calls are not guaranteed to be called in order? This would mean that samples could be written out of order and we would rely on the implementation of the MFT to deal with sorting the frames. Also as the completion of the tasks are never waited for there could `WriteSample` calls still queued while the `Finalize` call for the `sinkWriter` is executed.

By using `std::async` we get a `std::future` one can wait for to ensure that frames are written in order ([VideoEncoder.cpp:370](https://github.com/hermann-noll/MixedReality-SpectatorView/blob/09f62de01c1419c20dd7e1a1b1ea3ce47f4db7b2/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp#L370)) and that all frames are written before finalizing ([VideoEncoder.cpp:449](https://github.com/hermann-noll/MixedReality-SpectatorView/blob/09f62de01c1419c20dd7e1a1b1ea3ce47f4db7b2/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp#L449)).

### (Actual) hardware encoding

While trying to transfer the video texture directly to the video encoder without fetching its contents I stumbled upon [this line](https://github.com/microsoft/MixedReality-SpectatorView/blob/7796da6acb0ae41bed1b9e0e9d1c5c683b4b8374/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp#L75). It seems to me like by setting the result to an error code the attributes responsible for hardware encoding (and disabling throttling, lines 92-96) are never set? So mediafoundation could have chosen a software encoder all along.

### Encoding the video texture directly

As already stated by a TODO comment it would be great to encode the rendered video texture directly without temporarily fetching its content to RAM and back uploading back to GPU. Unfortunately this is a bit easier said than done.

Getting the textures to the video encoder is still easy enough, the texture should be copied into temporary one so the original can be modified while the older ones are encoded. The `VideoInput` class for hardware encoding needs a reference to the `ID3D11Device` though to create the texture (to not restrict the texture format of the input) and to trigger the GPU texture copy. The corresponding `IMFMediaBuffer` is created with `MFCreateDXGISurfaceBuffer` ([VideoEncoder.h:97](https://github.com/hermann-noll/MixedReality-SpectatorView/blob/09f62de01c1419c20dd7e1a1b1ea3ce47f4db7b2/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.h#L97)) which is then passed to the `sinkWriter` as normal.

In order to prepare the `sinkWriter` to accept textures it needs synchronized access to the `ID3D11Device` for which a `IMFDXGIDeviceManager` is required which needs to be set as attribute ([VideoEncoder.cpp:115](https://github.com/hermann-noll/MixedReality-SpectatorView/blob/09f62de01c1419c20dd7e1a1b1ea3ce47f4db7b2/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp#L115)). This manager was already created in the code but not passed. Also I think it is supposed to be used to synchronize the access to the device for the entire application? Which is not possible as we cannot control Unity's usage of the device. For an easy and working solution I opted to just move the synchronization to DirectX itself by using `ID3D10Multithread` ([VideoEncoder.cpp:61](https://github.com/hermann-noll/MixedReality-SpectatorView/blob/09f62de01c1419c20dd7e1a1b1ea3ce47f4db7b2/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp#L61)) but I guess a cleaner way would be to create a new headless `ID3D11Device` and use shared textures?

Also I could not get the encoding to work with the NV12 conversion, as `WriteSample` would return `MF_E_BUFFERTOOSMALL`. My guess would be that despite setting every input buffer and stating every buffer size as being  stride=4 (as RGBA), the encoder would try to internally copy the entire texture (which still is stride=4 after conversion) into a internally allocated buffer with stride=2 for NV12.

### Wrong sample time calculation?
I did not yet tackle the problem of some videos starting or ending with long freezes, I think the first one can be replicated by recording twice for a bit without leaving the play mode in between.

### Color conversions
I would guess that the RGB->NV12 conversion is not needed for hardware encoding anymore but would accelerate the CPU encoding (so turning the current logic around). The current state of this PR is that this conversion is disabled for all (with uncleaned parts still lying around)

### Audio side
Similar changes (especially the frame moving) can be applied to the audio side of the video encoder as well, but I wanted to discuss the more complex (and more performance-heavy) video changes with you first.

### DIFF
--- a/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
+++ b/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
@@ -347,8 +347,12 @@ void VideoEncoder::WriteVideo(std::unique_ptr<VideoEncoder::VideoInput> frame)
 #endif
     }
 
-    std::async(std::launch::async, [=, frame{ std::move(frame) }]() mutable
+    videoWriteFuture = std::async(std::launch::async, [=, frame{ std::move(frame) }, previousWriteFuture{ std::move(videoWriteFuture) }]() mutable
     {
+        if (previousWriteFuture.valid())
+        {
+            previousWriteFuture.wait();
+        }
         std::shared_lock<std::shared_mutex> lock(videoStateLock);
 
         HRESULT hr = E_PENDING;
--- a/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
+++ b/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
@@ -355,7 +355,7 @@ void VideoEncoder::WriteVideo(std::unique_ptr<VideoEncoder::VideoInput> frame)
         }
         std::shared_lock<std::shared_mutex> lock(videoStateLock);
 
-        HRESULT hr = E_PENDING;
+        HRESULT hr = S_OK;
         if (sinkWriter == NULL || !isRecording)
         {
             OutputDebugString(L"Must start recording before writing video frames.\n");
@@ -373,32 +373,7 @@ void VideoEncoder::WriteVideo(std::unique_ptr<VideoEncoder::VideoInput> frame)
 #endif
 
         IMFSample* pVideoSample = NULL;
-        IMFMediaBuffer* pVideoBuffer = NULL;
-        BYTE* pData = NULL;
-
-        // Create a new memory buffer.
-        hr = MFCreateMemoryBuffer(cbBuffer, &pVideoBuffer);
-
-        // Lock the buffer and copy the video frame to the buffer.
-        if (SUCCEEDED(hr)) { hr = pVideoBuffer->Lock(&pData, NULL, NULL); }
-
-        if (SUCCEEDED(hr))
-        {
-            //TODO: Can pVideoBuffer be created from an ID3D11Texture2D*?
-            hr = MFCopyImage(
-                pData,                      // Destination buffer.
-                cbWidth,                    // Destination stride.
-                frame->buffer,
-                cbWidth,                    // Source stride.
-                cbWidth,                    // Image width in bytes.
-                imageHeight                 // Image height in pixels.
-            );
-        }
-
-        if (pVideoBuffer)
-        {
-            pVideoBuffer->Unlock();
-        }
+        frame->Unlock();
 
 #if _DEBUG
 		{
@@ -408,11 +383,11 @@ void VideoEncoder::WriteVideo(std::unique_ptr<VideoEncoder::VideoInput> frame)
 #endif
 
         // Set the data length of the buffer.
-        if (SUCCEEDED(hr)) { hr = pVideoBuffer->SetCurrentLength(cbBuffer); }
+        if (SUCCEEDED(hr)) { hr = frame->mediaBuffer->SetCurrentLength(cbBuffer); }
 
         // Create a media sample and add the buffer to the sample.
         if (SUCCEEDED(hr)) { hr = MFCreateSample(&pVideoSample); }
-        if (SUCCEEDED(hr)) { hr = pVideoSample->AddBuffer(pVideoBuffer); }
+        if (SUCCEEDED(hr)) { hr = pVideoSample->AddBuffer(frame->mediaBuffer); }
 
         if (SUCCEEDED(hr)) { hr = pVideoSample->SetSampleTime(sampleTime); } //100-nanosecond units
         if (SUCCEEDED(hr)) { hr = pVideoSample->SetSampleDuration(frame->duration); } //100-nanosecond units
@@ -421,7 +396,6 @@ void VideoEncoder::WriteVideo(std::unique_ptr<VideoEncoder::VideoInput> frame)
         if (SUCCEEDED(hr)) { hr = sinkWriter->WriteSample(videoStreamIndex, pVideoSample); }
 
         SafeRelease(pVideoSample);
-        SafeRelease(pVideoBuffer);
 
         {
             std::shared_lock<std::shared_mutex>(videoInputPoolLock);
--- a/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
+++ b/src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
@@ -27,11 +27,7 @@ VideoEncoder::VideoEncoder(UINT frameWidth, UINT frameHeight, UINT frameStride, 
 #else
     inputFormat = MFVideoFormat_RGB32;
 #endif
-
-    for (int i = 0; i < NUM_VIDEO_BUFFERS; i++)
-    {
-        videoInputPool.push(std::make_unique<VideoInput>(frameHeight * frameStride));
-    }
+    inputFormat = MFVideoFormat_RGB32;
 }
 
 VideoEncoder::~VideoEncoder()
@@ -52,7 +48,21 @@ bool VideoEncoder::Initialize(ID3D11Device* device)
     if (deviceManager != nullptr)
     {
         OutputDebugString(L"Resetting device manager with graphics device.\n");
-        deviceManager->ResetDevice(device, resetToken);
+        hr = deviceManager->ResetDevice(device, resetToken);
+    }
+    for (int i = 0; i < NUM_VIDEO_BUFFERS; i++)
+    {
+        videoInputPool.push(std::make_unique<VideoInput>(device));
+    }
+
+    ID3D10Multithread* multithread;
+    device->QueryInterface(&multithread);
+    multithread->SetMultithreadProtected(TRUE);
+
+#else
+    for (int i = 0; i < NUM_VIDEO_BUFFERS; i++)
+    {
+        videoInputPool.push(std::make_unique<VideoInput>(frameHeight * frameStride));
     }
 #endif
 
@@ -79,7 +89,7 @@ void VideoEncoder::StartRecording(LPCWSTR videoPath, bool encodeAudio)
     prevVideoTime = INVALID_TIMESTAMP;
     prevAudioTime = INVALID_TIMESTAMP;
 
-    HRESULT hr = E_PENDING;
+    HRESULT hr = S_OK;
 
     sinkWriter = NULL;
     videoStreamIndex = MAXDWORD;
@@ -94,13 +104,14 @@ void VideoEncoder::StartRecording(LPCWSTR videoPath, bool encodeAudio)
 #endif
 
     IMFAttributes *attr = nullptr;
-    MFCreateAttributes(&attr, 3);
+    MFCreateAttributes(&attr, 4);
 
     if (SUCCEEDED(hr)) { hr = attr->SetUINT32(MF_SINK_WRITER_DISABLE_THROTTLING, TRUE); }
 
 #if HARDWARE_ENCODE_VIDEO
     if (SUCCEEDED(hr)) { hr = attr->SetUINT32(MF_READWRITE_ENABLE_HARDWARE_TRANSFORMS, true); }
     if (SUCCEEDED(hr)) { hr = attr->SetUINT32(MF_READWRITE_DISABLE_CONVERTERS, false); }
+    if (SUCCEEDED(hr)) { hr = attr->SetUnknown(MF_SINK_WRITER_D3D_MANAGER, deviceManager); }
 #endif
 
     hr = MFCreateSinkWriterFromURL(videoPath, NULL, attr, &sinkWriter);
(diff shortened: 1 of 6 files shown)


chrisfromwork:
There's a lot to digest in this review. In general, the recording stack could use some love and we are definitely open to these sorts of changes. I will try and provide a formal review/dive deep into these changes the beginning of next week.

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.h
```diff
@@ -38,33 +38,113 @@ class VideoEncoder
...
     void Update();
 
-private:
-    void WriteVideo(byte* buffer, LONGLONG timestamp, LONGLONG duration);
-    void WriteAudio(byte* buffer, int bufferSize, LONGLONG timestamp);
+    class VideoInputBase
+    {
+    public:
+        IMFMediaBuffer* mediaBuffer = nullptr;
+        LONGLONG timestamp = INVALID_TIMESTAMP;
```
chrisfromwork, line 53: If we have concerns around exposing IMFMediaBuffer to the CompositorInterface, we could also add a custom typedef here. I don't know that we will ever support this compositor on mac, but we could have a platform specific type as needed.

chrisfromwork: granted looking further below there are public constructors requiring references to other windows types (ID3D11Device, etc). I don't think there is a substantial difference to having the compositor interface know about media foundation types if it already knows about direct x types.

---
In reply to: [443729465](https://github.com/microsoft/MixedReality-SpectatorView/pull/395#discussion_r443729465) [](ancestors = 443729465)

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
```diff
@@ -21,10 +23,11 @@ VideoEncoder::VideoEncoder(UINT frameWidth, UINT frameHeight, UINT frameStride,
...
 {
 #if HARDWARE_ENCODE_VIDEO
-  inputFormat = MFVideoFormat_NV12;
+    inputFormat = MFVideoFormat_NV12;
 #else
-  inputFormat = MFVideoFormat_RGB32;
+    inputFormat = MFVideoFormat_RGB32;
 #endif
+    inputFormat = MFVideoFormat_RGB32;
 }
```
chrisfromwork, line 31: Does this change mean that hardware encoding succeeds using RGB32?

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
```diff
@@ -302,126 +325,91 @@ void VideoEncoder::WriteVideo(byte* buffer, LONGLONG timestamp, LONGLONG duratio
...
+#endif
 
-#if _DEBUG
-		{
-			std::wstring debugString = L"Writing Video Sample, SampleTime:" + std::to_wstring(sampleTime) + L", SampleDuration:" + std::to_wstring(duration) + L", BufferLength:" + std::to_wstring(cbBuffer) + L"\n";
-			OutputDebugString(debugString.data());
-		}
+#if !HARDWARE_ENCODE_VIDEO
+        frame->Unlock();
 #endif
```
chrisfromwork, line 392: Can we get a comment calling out when this is locked?

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
```diff
@@ -456,6 +444,11 @@ void VideoEncoder::StopRecording()
 
     concurrency::create_task([&]
     {
+        if (videoWriteFuture.valid())
+        {
+            videoWriteFuture.wait();
+            videoWriteFuture = {};
```
chrisfromwork, line 450: Its inconsistent, but i have seen an exception throw during this cleanup process for debug flavors of the compositor dll when running locally. I have been using the split channels recording option

chrisfromwork: specifically what looks like a stack overflow when setting this videoWriteFuture to an empty struct

---
In reply to: [443834338](https://github.com/microsoft/MixedReality-SpectatorView/pull/395#discussion_r443834338) [](ancestors = 443834338)

chrisfromwork:
I took a look at your review. In general, things seem to look good. In regards to your other follow up items, I have the following comments:

1. I also believe these incorrect sample time calculations may still exist. In testing your changes today I have not been able to repro the error. However, I am seeing a crash when recording for an extended period of time when stopping the compositor (inconsistent crash, but can somewhat consistently be reproed after multiple recording attempts). After figuring out this std::future cleanup, we may still see the same frame time errors. Do you know if this is something you have seen with hardware encoding, or is this something that may only exist for software encoding?

1. The folks that know the most about the origins of the nv12 codec conversions are out of office this week. I will follow up with them next week to understand whether the nv12 conversions are still needed.

1. Sounds good to look into the audio changes.

1. The HARDWARE_ENCODE_VIDEO macro may have not been properly used in the current master branch based on the bugs you found with these changes, but we have never suggested to external developers through documentation that the HARDWARE_ENCODE_VIDEO macro can be changed. I think we may be able to do away with software encoding to reduce our testing surface. I will reach out to some other folks internally that may have perspectives on whether they need the software encoding functionality. If you have preferences on keeping/removing the non-harware encoder video logic, let me know. :)

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
hermann-noll: For me it *only* worked with RGB32, I guess this is because the converted texture is still reported to D3D as being RGB. While it gets the information that the input is supposed to be NV12 with a therefore smaller buffer.
MF seems to be confused and writing samples failed with MF_E_BUFFERTOOSMALL.

src/SpectatorView.Native/SpectatorView.Compositor/Compositor/VideoEncoder.cpp
hermann-noll: I'll look into reproducing that exception, I haven't experienced any stack overflows before with the current changes.

hermann-noll: A very nice little bug, caused by `future` objects keeping the async lambda alive. By waiting on the previous future we create a chain of future objects where every element keeps the next one alive, growing rapidly over a long recording. When manually destroyed in line 450, this chain is recursively destructed causing the stack overflow.
I guess the really proper solution would be a more classic approach of a single task/thread consuming queued video frames, the chosen solution in cae0f9f just frees the future early, causing only the unprocessed frames to be in the chain.

chrisfromwork: that does sound quite nice. :)

In general, would you like me to test this out again locally so we can look into getting these changes in? Or would you prefer to continue looking into the TODO items?

hermann-noll: I would prefer a bit more work done to it, there are still some unclean parts to it. Nevertheless it would be great to know whether you guys have a dependency on the RGB->NV12 color conversion? (e.g. idk whether it works on Intel/AMD hardware encoder)

hermann-noll:
So I adapted the audio side to be more symmetric to the video side, nothing really new there, just two cycles less of allocation-copy-free for the audio frames.

While testing these changes by playing some music whilst recording I observed that in the recording the music was sped and chopped up. As this also occurred without my audio changes, this bug might be unrelated to this PR? I tracked it down to the sample time of the audio frames.
When the capture frame index is ignored and instead the sample time is calculated by adding up frame duration this bug went away, but this method is of course not very resilient to stutters in both audio and video.
Instead I drafted an estimation method to align DSP time and capture frame time which works alright, but I want to stress-test this one a bit more. Do you guys have any feedback regarding this method?

chrisfromwork:
I don't think anyone has a strong preference or suggestion for this time interpolation.

